PR #441, opened 2014-12-16
adds support for commented JSON configuration

ksatirli:
Our team likes to annotate JS(ON)-based configuration files and adding extra information to settings so "the next guy" knows why certain settings were made.

When possible, we use JS-based config files which support comments natively. In case we have to use JSON, we parse it with `cjson` so as to not break `JSON.parse` with inline-comments.

With this changeset, I added support for parsing commented JSON files (by means of `cjson`). I have also updated the tests.

tl;dr: added support for comments in nodemon.json

### DIFF
--- a/lib/config/load.js
+++ b/lib/config/load.js
@@ -1,6 +1,7 @@
 'use strict';
 var fs = require('fs'),
     path = require('path'),
+    cjson = require('cjson'),
     exists = fs.exists || path.exists,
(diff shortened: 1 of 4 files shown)


remy:
Breaks all [tests](https://travis-ci.org/remy/nodemon/jobs/44202151#L553). Not a good start...

PR #450, opened 2015-01-01
Fixed exec parser

### LINKED ISSUE #448: `\n` sequence in `--exec` argument needs a lot of escaping
Let's say I want to print something nice whenever nodemon triggers. Something really exciting, like:

    line1
    line2

In the shell, this would be:

    $ echo -e "line1\nline2"

In nodemon, I tried this:

    $ ./bin/nodemon.js -x 'echo -e "line1\nline2"'

(prints `line1line2`)

OK, so I guess I need to escape that backslash:

    $ ./bin/nodemon.js -x 'echo -e "line1\\nline2"'

(prints `line1line2`)

Third time's a charm?

    $ ./bin/nodemon.js -x 'echo -e "line1\\\nline2"'

(prints `line1line2`)

_4th_ time's a charm?

    $ ./bin/nodemon.js -x 'echo -e "line1\\\\nline2"'

success.

I am not sure what's causing this, but it feels to me like the amount of escaping required here is unreasonable.

For bonus points, if you try using double-quotes on the outside, and single quotes on the inside, you need a whopping _seven_ backslashes before it works:

    $ ./bin/nodemon.js -x "echo -e 'line1\\\\\\nline2'"

(prints `line1line2`), while:

    $ ./bin/nodemon.js -x "echo -e 'line1\\\\\\\nline2'"

works.

(all tested on current HEAD, 43ad1d82a2880bcc89913e58f3b44f0c26077d67)

remy:
- [x] ready for merge

It was losing quotes around arguments, and in fact only needs to capture the first element in the string as the exec, and the rest could be left as is (if I keep the `sh` method of execution).

Fixes #448

### DIFF
--- a/lib/config/command.js
+++ b/lib/config/command.js
@@ -2,13 +2,14 @@
 
 module.exports = command;
 
-function command(options) {
-  var executable = options.execOptions.exec,
+function command(settings) {
+  var options = settings.execOptions;
+  var executable = options.exec,
       args = [];
 
   // after "executable" go the exec args (like --debug, etc)
-  if (options.execOptions.execArgs) {
-    [].push.apply(args, options.execOptions.execArgs);
+  if (options.execArgs) {
+    [].push.apply(args, options.execArgs);
   }
 
   // then goes the user's script arguments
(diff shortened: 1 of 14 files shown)


remy:
:facepunch:

remy:
...but, as expected, this fucks up `nodemon --debug` and puts the `--debug` in the wrong position.

Adding tests and adding fix

remy:
:facepunch: :facepunch:

PR #452, opened 2015-01-05
changed --delay option: it now works like a throttle instead of with setTimeout

edi9999:
This changes the behaviour of the option --delay n

This means that as soon as one file changes, nodemon will restart the
script. The previous behavior was to wait for n seconds before
restarting the process. With throttle, the function gets called when the
first file changes. After that, all file changes in the interval [t,t+n]
are ignored

### DIFF
--- a/lib/monitor/watch.js
+++ b/lib/monitor/watch.js
@@ -262,16 +262,14 @@ function filterAndRestart(files) {
     config.lastStarted = Date.now();
 
     if (matched.result.length) {
-      if (config.options.delay>0)
-          {
-              if (throttleRestartBus===undefined)
-                  {
-                      throttleRestartBus=throttle(restartBus,config.options.delay,{trailing:false});
-                  }
-              return throttleRestartBus(matched);
-          }
+      if (config.options.delay > 0) {
+        if (throttleRestartBus === undefined) {
+          throttleRestartBus = throttle(restartBus, config.options.delay, {trailing:false});
+        }
+        return throttleRestartBus(matched);
+      }
       else {
-          return restartBus(matched);
+        return restartBus(matched);
       }
     }
   }
--- a/lib/monitor/watch.js
+++ b/lib/monitor/watch.js
@@ -261,12 +261,18 @@ function filterAndRestart(files) {
     // reset the last check so we're only looking at recently modified files
     config.lastStarted = Date.now();
 
-    if (throttleRestartBus===undefined)
-        {
-            throttleRestartBus=throttle(restartBus,config.options.delay,{trailing:false});
-        }
     if (matched.result.length) {
-      return throttleRestartBus(matched);
+      if (config.options.delay>0)
+          {
+              if (throttleRestartBus===undefined)
+                  {
+                      throttleRestartBus=throttle(restartBus,config.options.delay,{trailing:false});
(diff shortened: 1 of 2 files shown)


remy:
Do you think you can add a test to go with the code? I'm sure it looks good, but more test coverage is always good.

edi9999:
Can be merged now I think, I have fixed the special case where the delay is not set with commit 5b3fb19

edi9999:
Done

remy:
Awesome. I'm merging, and it'll ship to npm in the 1.3.0 release. I've still got a bit of work to get it fully complete, but excellent work - thanks!

edi9999:
Great :-)

remy:
Wow, I kinda got bit over this. The test did not work at all. It seemed to work in travis once, then it didn't at all. Now that I've debugged, I realise there were quite a few bugs in the test :'(

test/fork/watch-restart.test.js
```diff
@@ -86,6 +86,41 @@ describe('nodemon fork child restart', function () {
     });
   });
 
+  it('should happen only once if delay option is set', function (done) {
+    var restartCount = 0;
+    fs.writeFileSync(tmpjs, 'true;');
+
+    var p = run('--ext js --delay 1' + appjs, {
```
remy: `appjs` should have been `tmpjs`

test/fork/watch-restart.test.js
```diff
@@ -86,6 +86,41 @@ describe('nodemon fork child restart', function () {
...
+      error: function (data) {
+        p.send('quit');
+        cleanup(p, done, new Error(data));
+      }
+    });
+
+    p.on('message', function (event) {
+      if (event.type === 'start') {
+        setTimeout(function () {
```
remy: These two timeouts get set twice - but once the script restarts, it triggers the start event again, which causes this to reset. There's also often issues in Travis when touching files and nodemon picking up the change takes about 1 second, so these happening so closely to each other were hiding issues (along with the 1.5s timeout to check the status).

lib/monitor/watch.js
```diff
@@ -195,6 +195,51 @@ function ignoredFileTypesForFind(dir) {
   return ' \\( ' + paths.join(' -or ') + ' \\) -or ';
 }
 
+// Inspired by http://underscorejs.org/docs/underscore.html
+function throttle(func, wait, options) {
```
remy: This didn't actually work at all. A throttle will always let the first event through, and then *throttle* the rest. What it should have been was a [debounce](https://remysharp.com/2010/07/21/throttling-function-calls).

lib/monitor/watch.js
```diff
@@ -195,6 +195,51 @@ function ignoredFileTypesForFind(dir) {
...
+    result = func.apply(context, args);
+    if (!timeout) context = args = null;
+  };
+  return function() {
+    var now = new Date().getTime();
+    if (!previous && options.leading === false) previous = new Date().getTime();
+    var remaining = wait - (now - previous);
+    context = this;
+    args = arguments;
+    if (remaining <= 0 || remaining > wait) {
```
remy: This line always allows the first request through - which we didn't actually want for the delay param.

Issue #419, opened 2014-11-20
no automatic restart when using boot2docker docker volume


chrishiestand:
nodemon doesn't seem to restart node when using a docker volume, or in this case a boot2docker ( os x -> virtualbox -> docker) volume. nodemon starts okay, but then fails to restart node when a watched file has been modified in the shared OS X volume.

    root@760d5d2ff8cf:/src# nodemon --watch index.js 
    20 Nov 21:06:58 - [nodemon] v1.2.1
    20 Nov 21:06:58 - [nodemon] to restart at any time, enter `rs`
    20 Nov 21:06:58 - [nodemon] watching: index.js
    20 Nov 21:06:58 - [nodemon] starting `nodejs index.js`
    Running on http://localhost:8080
    request received
    #index.js changed and saved here but output does not change as it should
    request received
    request received

`nodemon -L` causes restarts at least once a second even when no files have changed - way too frequently.

`forever -w` seems to work just fine. That seems like the best workaround for now.

output of dump:

    root@b36ab2778133:/src# nodemon --dump
    20 Nov 22:08:27 - [nodemon] v1.2.1
    20 Nov 22:08:27 - [nodemon] to restart at any time, enter `rs`
    20 Nov 22:08:27 - [nodemon] watching: *.*
    --------------
    node: v0.10.25
    nodemon: v1.2.1
    command: node /usr/local/bin/nodemon --dump
    cwd: /src
    OS: linux x64
    --------------
    { run: false,
      system: 
       { cwd: '/src',
         useFind: false,
         useWatch: true,
         useWatchFile: false },
      required: false,
      dirs: [ '/src' ],
      timeout: 1000,
      options: 
       { scriptPosition: 0,
         dump: true,
         script: 'index.js',
         args: [],
         ignore: 
          [ '.git',
            'node_modules',
            'bower_components',
            '.sass-cache',
            re: /\.git|node_modules|bower_components|\.sass\-cache/ ],
         watch: [ '*.*', re: /.*\..*/ ],
         restartable: 'rs',
         execMap: { py: 'python', rb: 'ruby' },
         stdin: true,
         verbose: false,
         stdout: true,
         execOptions: 
          { script: 'index.js',
            exec: 'node',
            args: [],
            nodeArgs: undefined,
            ext: 'js',
            env: {},
            execArgs: [] },
         ext: 'js',
         monitor: 
          [ '*.*',
            '!.git',
            '!/src/node_modules/**/*',
            '!bower_components',
[...]

PR #453, opened 2015-01-05
Use a .nodemon-ping file to notify remote VM/container of file changes.

olalonde:
See issue #419 

When nodemon is used on a filesystem that doesn't support filesystem change notifications (e.g. Virtualbox shared folders which boot2docker uses to mount volumes for docker containers), nodemon doesn't detect file changes.

The solution I have come up with is to run nodemon on the host and toggle a `.nodemon-ping` file when changes are detected to the filesystem. 

```bash
nodemon --local
```

The client can listen for those events by polling `.nodemon-ping` and restarting the process when changes are detected:

```bash
nodemon --remote
```

It's a dirty hack, but it works.


remy:
Wow, okay, that's pretty cool. Not sure if it's ideal, but it's a start to crack a nasty problem.

One thing I wanted to ask initially (after reading code), I'm using Travis for tests, do you think/know if it's possible to automate these tests?

olalonde:
I will try to automate them. One issue I have found so far is that nodemon crashes if `ps` is not available in the Docker container. Also, the local nodemon should not be require to manage a process (it should just watch for filesystem changes). I will work on those issues today.

remy:
Take a look at what's currently sitting on the
fix/419-support-find-on-linux branch. It has a bunch of fixes for the Linux
env (and fixes the missing `ps` command).

I'm still not sure about having nodemon running on both environments, but
currently its the only viable solution available.

PR #459, opened 2015-01-08
Find on linux and detect container offset

### LINKED ISSUE #419: no automatic restart when using boot2docker docker volume

remy:
Closes #455
Closes #453
Fixes #419

This method creates a file and checks the stat to check of an offset between the date on the machine and the created time on the file system. Docker (in my tests) show that you can get the offset using this method.

Once we have the offset, we tell the user (once) and then we use that to find files that have been modified according to the offset (either looking in the last using -mmin or in the future using -newer + a touched file).

### DIFF
--- a/lib/monitor/watch.js
+++ b/lib/monitor/watch.js
@@ -6,7 +6,9 @@ var fs = require('fs'),
     bus = utils.bus,
     match = require('./match'),
     config = require('../config'),
+    offset = require('./offset'),
     childProcess = require('child_process'),
+    touch = require('touch'),
     exec = childProcess.exec,
     watched = [],
     throttleRestartBus,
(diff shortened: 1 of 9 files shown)


lib/monitor/watch.js
```diff
@@ -183,11 +196,11 @@ function ignoredFileTypesForFind(dir) {
     return '';
   }
   var paths = [];
-  if (dir.charAt(dir.length - 1) != '/') {
+  if (dir.charAt(dir.length - 1) !== '/') {
     dir += '/';
   }
   config.options.ignore.forEach(function (path) {
-    var pathIsDir = (fs.existsSync(dir + path) && fs.statSync(dir + path).isDirectory()) || (path.charAt(path.length - 1) == '/');
```
remy: These are just jshint fixes.

remy:
Merging because it also fixes the damn `--delay` option.

PR #503, opened 2015-02-11
Fix the pattern match for windows escaping

### LINKED ISSUE #497: Allow piping commands with execMap
I want to use nodemon to watch my .jsx files, so I've created `nodemon.json` like so:

```json
{
    "execMap": {
        "js": "node",
        "jsx": "jsx {{filename}} | node"
    },
    "ext": "js jsx",
    "ignore": [
        ".hg",
        "node_modules",
        ".idea"
    ],
    "verbose": true
}
```

But when I run nodemon I get:

> '\"jsx server.jsx | node\"' is not recognized as an internal or external command, operable program or batch file.

I guess it's escaping the command so it's treated as a single argument, or maybe it's not running in a shell so the pipe doesn't work? 

Either way, it would be nice if this worked.

As a workaround, I've created `jsx-node.bat` (I'm on Windows):

```bat
@echo off
jsx %1 | node
```

And changed `nodemon.json` to `"jsx": "jsx-node.bat"` instead. This seems to work, but the bat file has to be in the same directory as `nodemon.json` -- even paths like `bin/jsx-node.bat` don't work.

remy:
Fixes #497

Huge props to @xzyfer for solving it over [twitter](https://twitter.com/xzyfer/status/565525891678482432)

### DIFF
--- a/lib/monitor/run.js
+++ b/lib/monitor/run.js
@@ -58,12 +58,7 @@ function run(options) {
   // special logic for windows, as spaces in the paths need the path fragment
   // quoted, so it reads: c:\"Program Files"\nodejs\node.exe
   if (utils.isWindows && executable.indexOf(' ') !== -1) {
-    executable = executable.split('\\').map(function (part) {
-      if (part.indexOf(' ') !== -1) {
-        return '"' + part.replace(/"/g, '\\"') + '"';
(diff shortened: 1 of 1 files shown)


xzyfer:
<3

PR #515, opened 2015-03-02
Fix nodemon.json's 'quit' listener and add test.

STRML:
Previously, a `'quit'` listener in `nodemon.json` would not execute.

This is useful to have for cleanup in various cases. I was putting together a `nodemon.json` that was compatible with [naught](https://github.com/andrewrk/naught) in my case but I am sure there are plenty of others.

This patch simply iterates through all other `'quit'` listeners before calling `process.exit(0)`, to make sure they execute.

I have additionally removed some duplicated code in the SIGINT handler and added a test.

### DIFF
--- a/lib/monitor/run.js
+++ b/lib/monitor/run.js
@@ -271,7 +271,7 @@ run.kill = function (flag, callback) {
 };
 run.restart = noop;
 
-bus.on('quit', function () {
+bus.on('quit', function onQuit() {
   // remove event listener
   var exitTimer = null;
   var exit = function () {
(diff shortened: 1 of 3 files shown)


STRML:
I think the failing test is flaky. It appears unrelated. I ran it myself with `v0.8.28` and it passed.

lib/monitor/run.js
```diff
@@ -321,28 +327,11 @@ process.on('exit', function () {
 if (!utils.isWindows) {
   // usual suspect: ctrl+c exit
   process.once('SIGINT', function () {
-    var exitTimer = null;
-    var exit = function () {
```
remy: Why did all this block of code get removed? I'm not sure I can see why.

STRML: It's not obvious from the diff, but the removed code was functionally identical to the code in the `'quit'` handler. I wanted to remove the duplicated code for simplicity. See https://github.com/STRML/nodemon/blob/fixQuit/lib/monitor/run.js#L276

PR #529, opened 2015-03-26
Updated CoffeeScript link.

mortonfox:


remy:
Seems to been PR'ed twice, it's now in. Ta.

PR #535, opened 2015-04-16
Update travis for io.js

janders223:
Travis added support for io.js. This _should_ add that support if we want to check. Added the three latest io.js releases.


remy:
Included in #625.

PR #552, opened 2015-05-17
README specifies incorrect format for extensions

seldo:
The README specifies a different format for the list passed to `--ext` than the output of `--help` does, and it seems incorrect.


remy:
That was intentional, both formats work, or certainly the readme version is what I use day to day.

PR #566, opened 2015-06-12
Fixed restart "loop" due to VM clock out of sync with filesystem timestamps

diwo:
The "restart loop" issue can be caused by the system clock on the virtual machine becoming out of sync with the filesystem timestamps. The ```offset.js``` module is in place to address this issue, but the offset is only calculated during startup and does not take into account desync while nodemon is running. This can be replicated by putting the host machine to sleep (on OSX) for a few seconds while nodemon is running in the VM.

This patch addresses 2 issues:

1. The offset is updated every child process restart to prevent false positive change detection in the next ```find``` match due to incorrect delta values

2. The existing ```find``` delta value is quantized using a floor/ceiling function that produces incorrect positive delta values. For example, a ```1.9``` delta would be quantized to ```1``` and be used as the reference timestamp for ```find -newer```. This would find files changed after 1 second in the future instead of 1.9, returning more files than we want, including changes that have already been used to trigger previous finds.

### DIFF
--- a/lib/monitor/offset.js
+++ b/lib/monitor/offset.js
@@ -32,10 +32,14 @@ module.exports = function () {
   return offset;
 };
 
+module.exports.refresh = function() {
+  offset = null;
+};
+
 module.exports.pretty = function () {
   var date = new Date(offset);
   return two(date.getHours()) + 'h' + two(date.getMinutes()) + 'm' + two(date.getSeconds()) + 's';
-}
+};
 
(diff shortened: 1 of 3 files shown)


remy:
This PR has also broken mac support. From what I can tell, the `find -mtime` shouldn't be sub-second on a mac.

remy:
Superseded by #590.

PR #578, opened 2015-07-05
add option --no-startup to not run command on startup

### LINKED ISSUE #575: Is there a way to not run a command on startup but only when files change?
I'm using a cluster of workers in a webapp, and I would like nodemon to watch the files, and restart those workers (via a USR2 signal) whenever a change is made.

This is what I've done:

```bash
node some-app.js &
SERVER_PID=$!

nodemon -w modules -x "kill -s USR2 $SERVER_PID"
```

The only problem is, nodemon will automatically run the command on startup.  Is there a way to disable this, and only run when files are changed?

khoomeister:
Fixes #575


lib/cli/parse.js
```diff
@@ -191,6 +191,10 @@ function nodemonOption(options, arg, eatNext) {
     options.stdin = false;
   }
 
+  else if (arg === '--no-startup' || arg === '-S') {
+    options.runOnStartup = false;
```
remy: Sorry, just code reviewing now - this property should be named the same as the CLI flag, so it should be `noStartup` - but then I'm not sure if it should be true or false...

Should it be something like "execute on change", like `execOnChange` and `--exec-on-change`. Or `--no-immediate-start`...hmm, doesn't sound right either.

Overall the change is solid, tests included, but the flag naming needs to be consistent with the CLI and I'd like the naming to clear what it does.

Maybe use this issue to run through some options, then change the code once it's all settled?

khoomeister: OK sounds good... thinking out loud here:

```
-S, --no-startup ......... disable running script on startup.

-C, --exec-on-change ......... execute script on change only, not startup
-C, --on-change-only ......... execute script on change only, not startup

-S, --no-immediate-start ......... don't execute script on startup
-S, --no-start ......... don't execute script on startup
```

I like the `--exec-on-change` or `--on-change-only` variety, partly because it makes code easier to read (i.e. the double inversion of `noStart = false` is an unnecessary brain tax).

remy: Let's do `--on-change-only`

khoomeister: OK thanks, it's done & dusted :smiley:.

khoomeister:
*bump*... any chance of a merge & publish? :grin:

khoomeister:
Thanks! :smiley:

Issue #582, opened 2015-07-11
"watching 0 files" and watch detection is very slow


halt-hammerzeit:
Even when it's `watching *.*`.
What's the matter with this thing?

halt-hammerzeit:
When I delete my nodemon.json or make it an empty JSON object, it doesn't output "watching 0 files".

halt-hammerzeit:
If my nodemon.json is like this
```
{
    verbose: true
}
```

Then it outputs "watching 0 files"

halt-hammerzeit:
```
nodemon server/main.js --verbose --watch *.*
```

outputs "watching 0 files"

`ls` outputs a lot of files

halt-hammerzeit:
Still it says "files triggering change check: server\main.js"
So it's lying.

halt-hammerzeit:
And also the change detection is very slow.
It takes it about a second or two to notice the change.

halt-hammerzeit:
And it says `[nodemon] files triggering change check: webpack.config.js` even though I didn't tell it to watch that file: I told it to watch `server/**/*.js`

halt-hammerzeit:
Oh, it seems that after that it says `[nodemon] changes after filters (before/after): 1/0` so there is no match.
So watching patterns are not broken.
But the watching itself is very slow.

halt-hammerzeit:
And it still watches all the directories though I didn't tell it to do so.

halt-hammerzeit:
And why on earth does it watch the whole giant `config.dirs` folder and only when the watch triggers it checks the bloody regexp?

halt-hammerzeit:
```
   70        var cwd = process.cwd();
   71:       if (config.dirs.length === 0) {
   72:         config.dirs.unshift(cwd);
   73        }
```

dafuq?

halt-hammerzeit:
So this is a hidden secret "undocumented feature" and everybody suffers from that?

halt-hammerzeit:
So, the solution:

```
config.load = function (settings, ready) {
  reset();
  var config = this;
  load(settings, config.options, config, function (options) {
    config.options = options;

    if (options.dirs instanceof Array) {
      config.dirs = options.dirs;
    }
```

And that's it.
It's that simple.

Why?

halt-hammerzeit:
So, will you make the fix or do I have to fork the repo to is it with my projects?

halt-hammerzeit:
I'll go make a pull request

[...]

PR #583, opened 2015-07-11
Watch only the neccessary folders, not the whole huge folder

halt-hammerzeit:
See the corresponding issue: https://github.com/remy/nodemon/issues/582


halt-hammerzeit:
This pull request got messed up, will split it into two other ones.

PR #584, opened 2015-07-11
Fix excessive watching

halt-hammerzeit:
No need to watch the whole giant current folder recursively

### DIFF
--- a/lib/config/index.js
+++ b/lib/config/index.js
@@ -52,6 +52,10 @@ config.load = function (settings, ready) {
   load(settings, config.options, config, function (options) {
     config.options = options;
 
+    if (options.dirs instanceof Array) {
+      config.dirs = options.dirs;
+    }
+
     if (options.watch.length === 0) {
       // this is to catch when the watch is left blank
       options.watch.push('*.*');


remy:
Just to say I've seen your issue comments and PRs. I'm not in front of a machine where I can kick these tests to pass (I suspect they're messing up due to race conditions...). I'm hoping to be able to pick these up on Thursday and merge if they're good.

halt-hammerzeit:
@remy Ok, i've seen that some timings are failing (tests taking too long, I suppose).
If you need my help fixing them I can try to do that.
Cheers.

halt-hammerzeit:
@remy And no hurry, yes, surely it can wait.

halt-hammerzeit:
@remy 
By the way, I can see now why you might be watching the whole project folder.
This way you can detect more new files and new subfolders.
I don't know how much the performance penalty might be.
I guess it may be negligible.
But in some cases, assume something like a web downloader, it might get high (when the program constantly writes to disk).
If you think this kind of blind optimization is not worthy you can drop the pull request, I'm not an expert here.
But you better reflect this undocumented feature in the Readme because everyone thinks you just watch the regexped folders.

remy:
Hey. So, I've been trying to follow your train of thought from 11 days ago.

The undocumented feature is that yeah, it's watching all the files in the working directory and below. Since a new file could come along, it watches the directories, so nodemon is expected to be used on fairly shallow directories (and is supposed to exclude directories like `node_modules` etc).

The complexity comes when there's 3 platforms with different ways of doing things. `find` is being used in two different ways on Mac + Linux, and Windows uses `fs.watch` (and different methods depending on support). I *think* you're windows.

I'm not 100% sure if the issues you've seen, investigated and patched have completely done the trick or not, I'm hoping so, but please do let me know.

WRT this PR, I *think* the right thing to do is close it. I *think* it should be watching the current directory - but I'm not 100% sure what `options.dir` is (partly because I've been working on other projects for quite a while and I've forgotten what the detail is in this codebase).

halt-hammerzeit:
@remy ha ha, ok. yeah, i'm a windows fan. i'm having no issues about nodemon watching the whole codebase if it doesn't introduce significant performance penalty. at least this way you can track new files being added. "train of thought" is my thinking style (i stopped using drugs a long time ago). as for this pr we can surely close it since, as i've said in the other issue, the primary cause of the lag for me turned out to be the watch interval. that was interesting to have a chat with you though, you seem like a cool guy. cheers.

PR #591, opened 2015-07-21
Bump node-touch to 1.0.0

JoaoHenriquePereira:
[node-touch](https://github.com/isaacs/node-touch) has a bug where bin file is missing she-bang on version tagged v0.0.4 (automatically depended by nodemon) and impacting other modules depending on the bin file (such as libpq, bcrypt... on node-gyp rebuild cmd) thus breaking npm install.

Steps to reproduce
cd /tmp/
mkdir foo
cd foo
npm install touch@0.0.4
Install any lib that requires node-gyp rebuild
npm install libpq

### DIFF
--- a/package.json
+++ b/package.json
@@ -44,7 +44,7 @@
   "dependencies": {
     "minimatch": "~0.3.0",
     "ps-tree": "~0.0.3",
-    "touch": "~0.0.3",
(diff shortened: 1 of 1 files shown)


myndzi:
+1

rexxars:
+1

PR #681, opened 2015-10-12
Bump ps-tree dependency

cbas:
Solves toolchain warning about missing SPDX license in outdated
dependency ps-tree@0.0.3 -> event-stream@0.5.3 -> optimist@0.2.8.

### DIFF
--- a/package.json
+++ b/package.json
@@ -52,7 +52,7 @@
     "es6-promise": "3.0.2",
     "lodash.defaults": "3.1.2",
     "minimatch": "~0.3.0",
-    "ps-tree": "~0.0.3",
+    "ps-tree": "^1.0.1",
     "touch": "1.0.0",
     "undefsafe": "0.0.3",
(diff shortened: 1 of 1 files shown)


cbas:
Not sure what's happening here. Travis passes on Node 4.1.2 and Node 0.12 but fails on Node 0.10. However `npm test` fails on my local OSX 10.11 with Node 4.1.2. Some low level error in a C source file? :confounded: 

```
test/events/complete.test.js


  events should follow normal flow on user triggered change
    ✓ start (62ms)
    ✓ config:update
    ✓ exit (274ms)
    ✓ quit (734ms)
    ✓ stdout (50ms)
Assertion failed: (!(handle->flags & (UV_CLOSING | UV_CLOSED))), function uv_close, file ../deps/uv/src/unix/core.c, line 96.
sh: line 1: 14753 Abort trap: 6           ./node_modules/.bin/mocha --timeout 30000 $FILE
```

PR #693, opened 2015-10-19
fix: JSON extension not watched outside of coffeescript

### LINKED ISSUE #643: Detect changes in JSON files
Hi,

I'm using `nodemon` (v1.6.1) with Node (v4.1.0), it works perfectly but it doesn't detect changes in any ".json" files. Is this default behavior? How can it be changed?

Since in a server environment JSON files usually contain some settings, imho nodemon should restart by default when the content of a JSON file is changed.

Looking forward to hear your feedback, thanks!

pensierinmusica:
Fix a bug that caused `.json` file to not be watched outside of a coffeescript environment,
and improves the extension formatting regex.

Closes #643

### DIFF
--- a/lib/config/exec.js
+++ b/lib/config/exec.js
@@ -73,13 +73,8 @@ function exec(nodemonOptions, execMap) {
   var options = utils.clone(nodemonOptions || {});
   var script = path.basename(options.script || '');
   var scriptExt = path.extname(script).slice(1);
-  var extension = options.ext || scriptExt || 'js';
+  var extension = options.ext || scriptExt || 'js,json';
   var execDefined = !!options.exec;
-
-  // strip any leading periods int he extension
-  if (extension.indexOf('.') === 0) {
-    extension = extension.slice(1);
-  }
 
   // allows the user to simplify cli usage:
   // https://github.com/remy/nodemon/issues/195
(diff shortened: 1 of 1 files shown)


lib/config/exec.js
```diff
@@ -156,22 +151,11 @@ function exec(nodemonOptions, execMap) {
...
-      extension.indexOf('*.') !== -1) {
-
-    extension = extension.replace(/\s+/g, '|') // convert spaces to pipes
-      .replace(/,/g, '|') // convert commas to pipes
-      .split('|') // split on those pipes
-      .map(function (item) {
-        return item.replace(/^[\*\.]+/, ''); // remove "*."
-      }).join(','); // return regexp string like: js,jade
-  }
+  extension = extension.match(/\w+/g).join(',');
```
remy: This was the line I didn't like, but I've run some manual tests and it's okay, but it was this change that prompted me to *want* tests. I've manually checked the examples from the comments they're fine (because the regex is fine).

pensierinmusica:
When I changed that line I first made sure it worked fine on all the cases that you previously considered. I'm glad you had a chance to check it too. It should simplify the regex a bit and be more reliable too (as it covers more cases than the previous version, e.g. ".js, jade" becomes "js,jade" instead of "js,,jade").

So, do you think this can be merged now? Cheers  :)

remy:
I've been rerunning tests to get it to pass...

pensierinmusica:
Awesome  :)

PR #694, opened 2015-10-20
[docs] note on unix globs and --watch

timruffles:
I tried to use globbing with `--watch` and spent some frustrating time working out that I was being an idiot. Just popped a line in docs to help others.


remy:
:heart:

JounQin:
But why this is not supported? 😂

PR #695, opened 2015-10-20
fix: JSON extension not watched when "scriptExt" is defined

### LINKED ISSUE #643: Detect changes in JSON files

pensierinmusica:
Fix a bug that caused ".json" files to not be watched by default.

closes #643

### DIFF
--- a/lib/config/exec.js
+++ b/lib/config/exec.js
@@ -73,7 +73,7 @@ function exec(nodemonOptions, execMap) {
   var options = utils.clone(nodemonOptions || {});
   var script = path.basename(options.script || '');
   var scriptExt = path.extname(script).slice(1);
-  var extension = options.ext || scriptExt || 'js,json';
+  var extension = options.ext || scriptExt + ',json' || 'js,json';
   var execDefined = !!options.exec;
 
   // allows the user to simplify cli usage:


pensierinmusica:
It looks like "scriptExt" was avoiding "json" files to be watched.

Now it should be finally done once and for all. Sorry it took 3 PRs: the way config is set-up was a bit hard for me to fully grasp at first sight, since it's scattered in several places across the codebase.

Cheers  :)

pensierinmusica:
@remy the build fails with this error:
`nodemon exec should default to node`

I'm pretty sure though this has nothing to do with the bit of code that I touched in this PR.

Could you take a look please when you have a sec and merge it?

Thanks! Cheers  :)

remy:
The fact that it's failing on all three versions of node in the same place makes me think there's a bug. It might be in the test, but it's worth investigating.

PR #711, opened 2015-10-29
Watch js, json files by default

### LINKED ISSUE #710: Wrong default extension list
You can omit the file extension when using `node`, if you do so, right now only json files will watched. See 4c195df510eb97a6c86cc479f110a2f71e338062 that broke the proper behavior.

madbence:
In node the file extension is optional, so by default nodemon should watch
js and json files.

Closes #710


remy:
Is *this* change the cause of all the issues saying that nodemon isn't restarting? (going by your referencing this PR in issue #717).

madbence:
i've had the same issue, the workaround was https://github.com/madbence/spacegame/commit/5af62ca060d3343b4f3a7726c66b0dfa45064e4d

PR #718, opened 2015-11-02
Fixed a typo

dlmr:
Fixed a typo in `doc/option-parsing-logic.md`.


remy:
I should just remove this file, it's an original notes from the refactor about a year ago.

DLMR:
I would be fine with this file being removed, it does not give a end user of nodemon very much.

PR #724, opened 2015-11-10
feat: support --polling-interval command line arg

drtz:
Passes a polling interval to chokidar.watch() for legacy polling. Very useful for those of us running nodemon in a docker-machine image / VirtualBox VM.

### DIFF
--- a/lib/monitor/watch.js
+++ b/lib/monitor/watch.js
@@ -53,6 +53,7 @@ function watch() {
         ignored: ignored,
         persistent: true,
         usePolling: config.options.legacyWatch || false,
+        interval: config.options.pollingInterval,
       });
 
       watcher.ready = false;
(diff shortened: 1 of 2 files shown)


lib/monitor/watch.js
```diff
@@ -53,6 +53,7 @@ function watch() {
         ignored: ignored,
         persistent: true,
         usePolling: config.options.legacyWatch || false,
+        interval: config.options.pollingInterval,
```
remy: What happens if the user doesn't pass this value?

drtz: @remy chokidar falls back to it's default (100 ms) if interval is undefined. I can explicitly omit the option when undefined if you'd prefer that.

remy: No, it's fine if it's ignored if it's `undefined`. I'll merge up later tonight.

hedleysmith:
Any updates on this? It would be a great feature to have for minimising CPU usage in nodemon when running Nodemon on a mounted filesystem

PR #755, opened 2016-01-13
load: Add ability to specify custom config file

kentcdodds:
Specified via the `--config` flag

Hey @remy,
This PR is my way of asking you whether you're cool with doing something like this. I thought it'd be more helpful than an issue. I'm not sure whether this actually works, and I just made the change in the GitHub editor. If you're cool with this feature and you want me to, I'd be happy to make a legit PR with tests etc.

The reason for this is because I have a server that uses `nodemon.json` to configure how the server should run, but in addition, I have tasks (specifically to run [ava](http://npm.im/ava)) which require quite a different configuration. I'm sure I could do all of this with the CLI, but I'd rather not if I can help it :-)

Let me know what you think.

### DIFF
--- a/lib/config/load.js
+++ b/lib/config/load.js
@@ -165,7 +165,7 @@ function loadFile(options, config, dir, ready) {
     return callback({});
   }
 
-  var filename = path.join(dir, 'nodemon.json');
+  var filename = settings.config || path.join(dir, 'nodemon.json');
   fs.readFile(filename, 'utf8', function (err, data) {
     if (err) {
       return callback({});


kentcdodds:
Heh, this breaks tests. Like I said, wasn't sure whether this would actually work. Think of this as an issue with a bit more explicit (hopefully helpful) information :-)

lib/config/load.js
```diff
@@ -165,7 +165,7 @@ function loadFile(options, config, dir, ready) {
     return callback({});
   }
 
-  var filename = path.join(dir, 'nodemon.json');
+  var filename = settings.config || path.join(dir, 'nodemon.json');
```
remy: Shouldn't this just be `options.config`?

kentcdodds: Maybe so. I was mostly just filing this to see if you'd be interested and get some direction. Would you like an actual PR for this?

remy:
Yeah, I reckon it would be useful to others too to be able to configure where the file lives (I think it makes sense to only be controlled through the programmable API though, rather than the CLI options).

knksmith57:
Heya @kentcdodds + @remy!

Don't mean to step on any toes, but I came across the need for this feature today and decided to pick up the PR and flesh out the patch. Just opened up #855 (what are the odds!) to do so.

Looking for feedback and to see if we could (kindly) get this thing rolling along.

Thanks for spiking on the original feature, @kentcdodds. I'd let you squash and take credit for the patch, but I think your commit graph is colorful enough already 😉

PR #778, opened 2016-02-08
Fix buildup of ignored option for Chokidar

novemberborn:
I found some bugs in how dotfiles are not ignored if explicitly watched. This code was introduced in #634. 

If the initial `ignored` is undefined there's a crash. Otherwise the dotfile pattern is appended to the expression, invalidating its last part. E.g. `/\.git|node_modules|bower_components|\.sass\-cache/` becomes `/\.git|node_modules|bower_components|\.sass\-cache[\/\\]\./`. It should have been `/\.git|node_modules|bower_components|\.sass\-cache|[\/\\]\./`

The fix is to construct an array again, as was done in the original code. An undefined `ignored` will be ignored by Chokidar (no pun intended), and there are no regex concatenation problems.

I did a quick scan of the issues list. I believe this may be the root cause of #762 and #723.

Test output before the fix commit:

```
$(npm bin)/mocha --ui bdd test/monitor/match.test.js

[...]

watcher
    1) should not crash if ignoreRoot is an empty array
    2) should not match a dotfile unless explicitly asked to
    ✓ should match a dotfile if explicitly asked to


  25 passing (274ms)
  2 failing

  1) watcher should not crash if ignoreRoot is an empty array:
     TypeError: Cannot read property 'source' of undefined
      at lib/monitor/watch.js:49:37
      at lib$es6$promise$$internal$$initializePromise (node_modules/es6-promise/dist/es6-promise.js:373:9)
      at new lib$es6$promise$promise$$Promise (node_modules/es6-promise/dist/es6-promise.js:664:9)
      at lib/monitor/watch.js:43:19
      at Array.forEach (native)
      at Object.watch (lib/monitor/watch.js:42:8)
      at test/monitor/match.test.js:355:20
      at lib/config/index.js:79:7
      at lib$es6$promise$$internal$$tryCatch (node_modules/es6-promise/dist/es6-promise.js:326:16)
      at lib$es6$promise$$internal$$invokeCallback (node_modules/es6-promise/dist/es6-promise.js:338:17)
      at lib$es6$promise$$internal$$publish (node_modules/es6-promise/dist/es6-promise.js:309:11)
      at lib$es6$promise$asap$$flush (node_modules/es6-promise/dist/es6-promise.js:120:9)

  2) watcher should not match a dotfile unless explicitly asked to:

      AssertionError: should not contain .dotfile
      + expected - actual

      -1
      +0

      at test/monitor/match.test.js:372:20
[...]

PR #779, opened 2016-02-08
Get default ignoreRoot from ignore-by-default

novemberborn:
Hello!

I'm [adding `--watch` support to AVA](https://github.com/sindresorhus/ava/pull/502). Like nodemon it uses Chokidar to watch for changes, and like nodemon it should ignore non-code directories. Initially I copied the `ignoreRoot` config from here but it seemed like the list of directories to ignore should be in a package of its own. That way multiple projects can share good defaults :100: 

This PR switches the default `ignoreRoot` to a list obtained from the [`ignore-by-default` package](https://github.com/novemberborn/ignore-by-default). Compared to the original list in nodemon this adds `.nyc_output` and `coverage` directories.

I marked this as a `refactor`, but maybe it should be a `feature` instead. If so maybe the `ignore-by-default` dependency should be pinned to an exact release so changes won't impact nodemon.

This requires #778 to land first, otherwise that bug causes test breakage in this PR.

### DIFF
--- a/lib/config/defaults.js
+++ b/lib/config/defaults.js
@@ -1,3 +1,5 @@
+var ignoreRoot = require('ignore-by-default').directories()
+
 // default options for config.options
 module.exports = {
   restartable: 'rs',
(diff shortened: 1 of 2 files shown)


remy:
@novemberborn just wanted to drop a note - I'm aware of your two PRs <3 I need to grab some time to get this merged in, and work out why the travis is failing (though it's likely to be the sheer number and scary race conditions in the tests!).

novemberborn:
@remy thanks for the update!

remy:
merged into master.

Issue #769, opened 2016-02-02
Should support unicode extensions


alice0meta:
I tried `nodemon -e ζ ./server.js "$@"`; https://github.com/remy/nodemon/blob/master/lib/config/exec.js#L158 threw an error (currently https://github.com/remy/nodemon/blob/3480f37aab325b6ae4e17058078742d586724860/lib/config/exec.js#L158 ); I changed that line to `extension = extension.match(/[\wζ]+/g).join(',');` and it works fine now. Please change the regex to be something more permissive, like `/[^,.\s]+/g`.

remy:
Can you send a PR with a test passing. Thanks.

alice0meta:
I haven't interacted with github much and don't know how to do that. Please change the regex from `/[\wζ]+/g` to `/[^,.\s]+/g` or something else that’s blacklist-based instead of whitelist-based.

alice0meta:
@remy explained that he WONTFIX

alice0meta:
It's unclear if this should be kept open. I apologize to anyone else who has this issue; I was unable to even get clarification if the maintainers of nodemon care about it.

remy:
<img width="794" alt="screen shot 2016-02-25 at 09 59 13" src="https://cloud.githubusercontent.com/assets/13700/13316128/73899554-dba6-11e5-948e-201b7a4c61c3.png">

alice0meta:
Thank you. I'll check back later to see if this seven-character fix has been implemented.

remy:
https://github.com/remy/nodemon/issues/769#issuecomment-178439145

deckar01:
Fixed in https://github.com/remy/nodemon/pull/789.

Thanks @remy for working on the test.

Thanks @alice0meta for finding a solution to this issue.

PR #787, opened 2016-02-25
fix: unicode extensions not supported

alice0meta:


alice0meta:
Afaict this actually works just fine; one of the tests was faulty.

remy:
Can you a) add a test for this change to support the issue, b) rebase the commit to use the contributing format?

Thanks, otherwise looks good. I've got a few PRs that I need to merge and need to look at why these tests are causing issues (and randomly failing).

alice0meta:
I don't know how to do that. Please just use this. It's a seven-character fix.

remy:
If you don't rebase it won't be released. Not because I don't want to
release it but because the release is based on the commit message.

alice0meta:
Can you please just make the edit yourself then? I've tried. I don't know how to do this. It's only a seven-character fix.

remy:
Or in fact, since you've only a single commit you should be able to amend
your commit and push force it back.

Note, if don't add a test, it's fine in this case, but if the code is later
refactored the bug could be reintroduced. A test on this would protect the
fix in the future, that's all.

alice0meta:
I understand you're not willing to fix this bug. I've closed the issue.

remy:
Tell me what I'm supposed to test against, and if I have time, I'll redo
the commit. Sorry, I'm busy with paid work too :)

I appreciate the PR but I'm just trying to help you get it live.

alice0meta:
The fix - which replaces a simple regex with a slightly different simple regex - is precisely defined https://github.com/remy/nodemon/issues/769 . I understand that you're not willing to implement it.

remy:
"Not willing" you're kidding right? I tried to help you, but this isn't a
paid project for me, so I commit time when *I want*.

Have a read: http://alistapart.com/article/the-high-price-of-free

I'm willing to help you, but I'll only redo this work when I have time.

alice0meta:
I understand that you prefer spending time arguing on the internet to fixing simple bugs. I don't blame you. Arguing on the internet is fun. I understand that nodemon does not and will not support languages other than English.

remy:
I've tried to help you to get this live. You refuse to use the contributing guide. This change literally won't go live because the commit message is wrong. I'm not arguing, I'm replying on my phone, *trying* to help you get this live.

alice0meta:
I haven't done this before. I don't know how to use this complicated system. I understand that the maintainers of nodemon do not and will not support languages other than English.

remy:
Right. How about we start again. Let's be friends. I am the maintainer of nodemon.

I want to help you. Honestly.

Your commit is great because it adds support (I believe) for non English characters in filenames. :+1:

So, the problems *we* need to fix are:

1. The commit message won't trigger a build
2. We don't have any tests protecting your change from regression

You don't know how to do these things. Fair enough, if you'll let me teach you, then *we* can get your fix live, and you'll have made nodemon better for non-english filenames.

How does that sound?

alice0meta:
That sounds like it would take much more time for both of us than you typing seven characters. You've typed a lot more than seven characters already, and that would involve you typing **many** more. I don't think either of us want that.

remy:
I'm afraid a 7 character change will always be way less than the sum character of it going live.

I'm genuinely happy to help you get this in. But it needs both of us to give a little more time. And this way, you'll have a few more tools to your skills to make other projects better in the future.


Your call.

alice0meta:
I understand that the maintainers of nodemon do not and will not support languages other than English.

remy:
Not sure if that repeat message was a mistake or intentional, but to clarify, I am the maintainer and I will support non-english users. 

Do you want to help achieve that? https://github.com/remy/nodemon/pull/787#issuecomment-188674117

Or not? It's you call.

alice0meta:
Thank you for confirming that you will support non-english users. I'll check back in a week to see if the regex has been replaced.

remy:
So I'm clear, are you interested in my helping you, as per https://github.com/remy/nodemon/pull/787#issuecomment-188672958 or are you saying you not interested and you'd rather someone else do the work instead?

alice0meta:
Incorrect. I've done the work to locate a bug and find the precise fix. I'm interested in the maintainers of nodemon doing the work to interface with their own system to publish the fix.

remy:
Maintainers? You mean, me.

So, you're done, and you want me to finish your work.

And, no, you're not interested in me helping you.

Loud and clear.

alice0meta:
Correct. I've done most of the work, and I'd like you to use your domain-specific knowledge to do the last few keystrokes much more efficiently than I could.

Will nodemon support languages other than English, or won't it? If it won't, I'd like to know already, so that I can abandon it. If it will, then I have nothing else to do here.

remy:
See replies above.

remy:
Locking conversation now because you seem to have dropped into an infinite loop.

PR #822, opened 2016-04-23
Add child:start event

akesser:
The childStarted event gets emited when the child logs a special string, default is "nodemon::started".

I use gulp-nodemon with browser-sync. When nodemon reloads the server, browser-sync should also reload. The current approach is to listen for restart or start events and using setTimeout to call browser-sync's reload function. The timeout is needed, because the reload must wait for the server to properly start up, if the server is not ready, the browsers try to reload and timeout.

My approach to solve this is to call  

``` javascript
console.log("nodemon::started")
```

when the server is ready and than to trigger a "childStarted" event to avoid the setTimeout call.


akesser:
The failing check is due to a timeout in a node 0.10 environment (the same error as in pull request # 823) and not due to changes in the code

remy:
I'll take a look but I didn't see any tests with the change you pushed
either (if you could add these as it's part of the contributing spec).
Thanks.

akesser:
Again this timeout error with the quit event, but this time with node 5

remy:
I've got this test passing (sorry for the delay). One thing though, can you change the event name from `nodemon::started` to `start:child` to be consistent with the existing event names?

Thanks - hopefully I'll be a bit faster on kicking the tests to get them to pass.

akesser:
I think it is fine to call the event `start:child`, but in my opinion the string that triggers this event should contain `nodemon`, for example `nodemon:start:child`. With this approach one gives a hint to the programmer that looks only at the app.js file, so he knows that this string is needed for nodemon.

remy:
Ah, I was referring to the event that fires from nodemon out to the wrapping service (like gulp).

Also agree to make the default user text be `nodemon:start:child` 👍

akesser:
changed the name of the event and the default user text to be `nodemon:start:child`

akesser:
To be honest, I do not know how to fulfill the ci/sematic-release check

remy:
You need to rebase your commits (replying on mobile, so I don't have any examples to hand).

akesser:
I rebased the code. The error in the ci check is against due to a timeout with node 0.10 and node 4 in the `test/events/complete.test.js` file

ramasilveyra:
Great work @akesser! Can't wait for this 😄

remy:
The timeouts are set to 30 seconds, so if they're timing out, it means the test is failing in this case. Have you tried to replicate the running tests locally?

akesser:
Running `npm test` on my Mac with node 6 works just fine.

akesser:
Any progress on this?

corysimmons:
Bumping. Would love to use Nodemon and BrowserSync together (preferably from BrowserSync CLI).

jvanveen:
Would love to see this. Running into the same issue with the timeout and livereload in gulp

corysimmons:
@jvanveen https://github.com/corysimmons/chokidar-cli-infanticide kills all children processes. Not sure if it's helpful to you in a Gulp pipeline.

Note, that pkg doesn't help with BrowserSync. For my part, I think I had a misunderstanding of what Koa was doing with HTTP responses. Koa is just returning something, not creating an injection point (an .html file) for BrowserSync to inject its little script that enables live reload.

https://gist.github.com/corysimmons/c6b31bc7d5d9d970d884175a293d8a0d is a pretty simple/straightforward solution for what I wanted (i.e. being able to update routes and see their data updated in the browser in realtime).

I suspect I'm just doing that off-topic newb rambling thing, but wanted to share my experience for anyone else looking for similar functionality to what I was trying to do. I'm positive there's a better way to do it than relying on BrowserSync's delay, but I can't be bothered right now.

remy:
I'm closing this due to inactive. I _do_ think there's something here, but I'd like to have a very clear understanding of the use case. Happy to reopen, but I know there's a bunch of merge conflicts right now.

Issue #856, opened 2016-06-16
Restarting a process with a java process causes a java core dump


Sticksman:
When running nodemon on a file that spawns a java process, restarting results in a core dump. 

This is because sending SIGUSR2 to Java causes a segfault. Digging a bit it's because SIGUSR2 === kill signal 31. According to this: http://journal.thobe.org/2013/02/jvms-and-kill-signals.html, Java handles this in the same way as a segfault.

Issue #745, opened 2016-01-05
"[nodemon] clean exit" when using `express'` `res.render`


derenio:
When I run my service with the `nodemon 1.8.1`, with:
```
nodemon --debug --delay 0.5 -i 'public/' -i 'client/' -i 'node_modules/' main.coffee
```
The server works as expected until it needs to execute a `res.render` function, example:

```coffee
...
    view: (req, res) ->
        context =
            some: 'data'
        res.render 'view', context

...
```
with an error:
```
[nodemon] clean exit - waiting for changes before restart
```
It does work normally when I run it with `coffee main.coffee` and it worked until I've updated `nodemon` to version `1.8.1`.

I'm writing it in `coffee-script` and use `jade` as a rendering engine, running on `nodejs` 5.3.0.

derenio:
Turns out it's an issue with the newest `nodejs` versions (starting from 5.2.0).

https://github.com/nodejs/node/issues/4261

P.S. I've removed the logging as in the provided example and my code started to work. Turns out you shouldn't even remove the `console.log` from ones' crashing example...

PR #857, opened 2016-06-16
Kill java processes with SIGUSR1

Sticksman:
Fixes https://github.com/remy/nodemon/issues/856


Sticksman:
Note, I've tested this on my box, but this is difficult to test without adding a bunch of dependencies to this project.

krevels:
This is may no longer necessary since #745 makes the signal configurable

PR #871, opened 2016-07-06
Include TypeScript files by default

arosequist:
If CoffeeScript files are included by default, it seems reasonable to also include TypeScript files by default.

---

This change is [Reviewable](https://reviewable.io/reviews/remy/nodemon/871)

### DIFF
--- a/README.md
+++ b/README.md
@@ -136,7 +136,7 @@ Don't use unix globbing to pass multiple directories, e.g `--watch ./lib/*`, it 
 
 ## Specifying extension watch list
 
-By default, nodemon looks for files with the `.js`, `.coffee`, `.litcoffee`, and `.json` extensions. If you use the `--exec` option and monitor `app.py` nodemon will monitor files with the extension of `.py`. However, you can specify your own list with the `-e` (or `--ext`) switch like so:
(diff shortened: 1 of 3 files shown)


remy:
I'm not sure I agree…(only because there's been no demand for this), but I think it does raise the valid question: should coffee script still be included by default? I'm not sure it should anymore.

rjmunro:
When coffee script files are included, how do they then get compiled into what is re-run?

For typescript, I run `tsc -w` to watch the typescript and compile it to js, then run nodemon on that output. I'm using concurrently to run both at once as follows:

    concurrently 'tsc -w' 'nodemon --watch "build" build/server.js'

remy:
Actually, I'm more likely to strip out coffee script support. It was added
nearly 6 years ago, and coffee script doesn't have the same importance that
it had back then (due to es6).

You can easily make typescript control your starting using the exec alias
config in your global nodemon. I'd be happy to take a FAQ update for this.

remy:
Sorry, closing this. I can't remember where it went, and can only see doc changes (though I'm not familiar with TS so it's possible I'm missing something).

Issue #878, opened 2016-07-19
Allow .env file to be specified via cli flag


mettjus:
It would be very useful if a path to a `.env` file could be specified without the need of creating an additional `nodemon.json` to inject environment variables. Right now there are a number of workarounds (eg: using Foreman or running commands such as `env $(cat .env) nodemon app.js`) to achieve this but they are all less than optimal.
Something like `nodemon --env-file .env app.js` would be great!

remy:
Why not use npmjs.com/dotenv to read this enviroment file instead?

mettjus:
You mean inside the app? I normally use .env as a placeholder for environment variables I later inject in my docker containers. Another way I sometimes work around the impossibility of specifying the env file via flag is using dotenv inside a guplfile for dev only purposes, but of course avoiding it would be much more comfortable :)

yanneves:
This could become too convoluted. Also, you may prefer to use [dotenv-safe](https://github.com/rolodato/dotenv-safe), which would warrant another argument like `--env-sample-file`.

To make use of dotenv right now in Nodemon, I'd suggest using:
```bash
nodemon --exec "node -r dotenv/config" script.js
```

AFAIK, there's no config in Nodemon for passing command-line options to Node, along the lines of:
```bash
node [options] script [arguments]
```

Maybe that's something that should be supported (or already is)?

mettjus:
The `nodemon --exec "node -r dotenv/config" script.js` trick is a valid workaround, IMHO. Thanks @yanneves.

yanneves:
No problem!

maxfrigge:
In case anyone else needs it :)

This works: `nodemon --exec 'node -r dotenv/config' dist/server.js dotenv_config_path=config.env`

j0hnm4r5:
And if you were also using something like esm, you'd do it like this: `nodemon --exec 'node -r esm -r dotenv/config' index.js`

09wattry:
I was able to do this by providing the path option to my .env file in my execution script. I used a relative path, which worked, or you could use the path resolve() function to get an absolute path.

```javascript
const { resolve } = require('path');
const envPath = resolve(process.cwd(), '../.env');

require('dotenv').config({ path: envPath });
```

PR #879, opened 2016-07-20
feat: support --env-file command line arg

mettjus:
Refers to issue #878. Should be working fine. Must I create tests for this one?

### DIFF
--- a/lib/cli/parse.js
+++ b/lib/cli/parse.js
@@ -197,6 +197,11 @@ function nodemonOption(options, arg, eatNext) {
     // grunt-nodemon - we're doing this early because it will affect where the
     // user script is searched for.
     process.chdir(path.resolve(options.cwd));
+  } else
+
+  if (arg === '--env-file') {
+    options.env_file = path.resolve(eatNext());
+    require('dotenv').load(options.env_file);
   } else {
 
     // this means we didn't match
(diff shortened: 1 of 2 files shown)


remy:
I'm not super keen on this adding new arguments.

I do understand the need for injecting environment values (and although this can be done in `nodemon.json` I appreciate it's not straight forward).

On tests: yes, you would. But I'm wondering if there's something smarter that can be done here. Perhaps pointing the `env` property to the file and supporting on the CLI (going against my rule of not adding new CLI args).

mettjus:
> On tests: yes, you would.

Ok. I will try to figure them out.

> Perhaps pointing the env property to the file and supporting on the CLI

If you mean add the `env` property to nodemon.json and point it to the .env file I think this would be way more comfortable than the current behaviour but still less than optimal. I say this because I find nodemon incredibly useful to quickly start watching some files without the need of configuration and requiring to create a nodemon.json just to specify the .env file location would look useless setup overhead to me.

remy:
>  say this because I find nodemon incredibly useful to quickly start watching some files without the need of configuration and requiring to create a nodemon.json

Absolutely, and part of the design policy (in the application) is that *every* `nodemon.json` config is available via the CLI, so this is where I *would* add the `--env` support on the CLI (since it's currently missing).

remy:
Late to follow up on this, but I'm going to close. Thinking behind it is that the ENV loading should be outside of nodemon (or inside the target).

PR #922, opened 2016-10-15
feat: change ignore rules to match whole name

### LINKED ISSUE #916: folders with suffix .git not detecting changes
Hi @remy 

I use .git suffix for my git projects. When I try to use nodemon on lets say folder `Projectxx.git`; they changes won't get detected.

Thx

mattlyons0:
Previously ignore rules would be fuzzy compared (a rule of '.git' would match 'myproject.git') and have unintended consequences.
Fixes #916

### DIFF
--- a/lib/rules/add.js
+++ b/lib/rules/add.js
@@ -82,6 +82,9 @@ function add(rules, which, rule) {
       return rule.replace(reEscapeChars, '\\$&')
                  .replace(reAsterisk, '.*');
     }).join('|');
+    // Only match if a '\', '/' or nothing comes before item
+    // Additionally only match if a '\', '/' or nothing comes after the item
+    re = '(\\/|\\\\|^)(' + re + ')($|\\/|\\\\)';
 
     // used for the directory matching
     rules[which].re = new RegExp(re);


mattlyons0:
This may break some configurations where users were taking advantage of the fact that the ignore rules matched fuzzy. This affects both the default configuration of ignore rules as well as user specified.

See my comment on #916 for why I think the old behavior should be changed. The short version is that if the full path of a project folder had a folder with '.git', 'coverage' or any of the folders in the `ignore-by-default` package as a substring of the folder name the entire directory tree up from that folder would be ignored. Folders like 'mycoverageapplication' would be ignored entirely with the old behavior and default settings. The same principle applies to files. There could still be an issue if somewhere in the file tree a user puts their code in a folder named explicitly 'coverage' or 'node_modules' however I don't think that happens very often and would be more obvious than previous behaviors matching.

rjmunro:
Does this apply to all ignore rules or only the default ones?

mattlyons0:
All ignore rules, user configured and ignoreRoot.

remy:
Is it possible to add some tests directly to this function, at least to show it does what it says on the tin?

remy:
I know this PR is old, and I'm going to close it for now, since I suspect you've possibly moved on. But I'd be very happy to re-open and merge if a test can be added to confirm the expected functionality.

PR #946, opened 2016-12-01
Exit with code 1 on crash if --exitcrash

bdentino:
I'm using nodemon in a docker container with a restart policy configured to restart the container in case it crashes. If my app crashes, nodemon exits with code 0 and thus docker assumes it was a graceful exit, so it does not restart my container. I don't see the harm in changing the exit code here to reflect an unexpected termination, given that this path is only exercised in the event of a crash.


manixx:
I have the same problem. Looking forward to get this merged. :)

PR #949, opened 2016-12-06
fix: avoid changing parent's current working directory

rlisagor:
Currently, if you use nodemon as a library with the `cwd` argument, nodemon will change the current working directory of the parent process. In many cases, this is undesirable. This change uses the `cwd` argument to `child_process.spawn` instead.

Thoughts / comments / requests welcome.


remy:
Super-super late to this PR, but that particular flag is intended to do exactly that - to _change_ to the `cwd` provided so that nodemon is running in the root of that directory.

Issue #901, opened 2016-08-31
Running nodemon EPERM: operation not permitted, rename on configstore even if I have opt out [windows]


manuel-di-iorio:
Error thrown even if I had `"optOut": true`

```
Error: EPERM: operation not permitted, rename 'C:\Users\User\.config\configstore\update-notifier-nodemon.json.30118632' -> 'C:\Users\User\.config\configstore\update-notifier-nodemon.json'
   at Error (native)
   at Object.fs.renameSync (fs.js:681:18)
   at Function.writeFileSync [as sync]
   etc...
```

amelon:
I have the same issue.
Is there any workaround?

amelon:
related issue on update-notifier
https://github.com/yeoman/update-notifier/issues/68

manuel-di-iorio:
@amelon try to delay the startup of two `nodemon` concurrently processes with the `timeout` command bash (both on windows and linux)

PR #953, opened 2016-12-13
Update update-notifier dependency

ysangkok:
This is useful because update-notifier 0.5 uses latest-version 1.x which has [return statements outside functions](https://github.com/sindresorhus/latest-version/blob/v1.0.0/cli.js#L24) which prevents it from being used with Babel (it is a syntax error).


paulpatarinski:
Updating notifier also fixes #901

remy:
bump coming in another PR. But thank you.

PR #957, opened 2016-12-29
fix: stripping the SIG prefix for POSIX compliance

### LINKED ISSUE #956: the SIG prefix is not valid according to the POSIX standard
Nodemon fails to reload the application it supervises.

When nodemon detects a change in the watched directory it tries to signal the child using the following (let's assume the child's pid is 123):
`/bin/sh -c kill -s SIGUSR2 123`
/bin/sh under most of the linux distros is attached like this:
`/bin/sh -> /bin/bash`
which triggers in strict POSIX compliance. A POSIX compliant shell won't understand the SIG  prefixes. Following this discussion http://unix.stackexchange.com/questions/199360/kill-sigstop-invalid-signal-specification-error-in-bash-script I successfully reproduced the behavior.
You can either start bash using the --posix option or start /bin/sh
Then `kill -s SIGTERM 123 or kill -s SIGUSR 123 `will yield:
`bash: kill: SIGTERM: invalid signal specification`
or 
`/bin/sh: 2: kill: invalid signal number or name: SIGTERM`
respectively. Similar things happen when the ps utility is installed.

The problem current codebase has is that it tries to pass the SIG* constants when constructing the signaling command and it tries to use a strictly POSIX shell at the same time.

I'm trying to run nodemon 1.11.0 under amazon linux and debian jessie.

knight:
changes which fix #956

### DIFF
--- a/lib/monitor/run.js
+++ b/lib/monitor/run.js
@@ -276,9 +276,16 @@ function kill(child, signal, callback) {
       //  an array of PIDs that have spawned under nodemon, and we send each the
       //   configured signal (defaul: SIGUSR2) signal, which fixes #335
       psTree(child.pid, function (err, kids) {
-        spawn('kill', ['-s', stripPrefix(signal), child.pid].concat(kids.map(function (p) {
-          return p.PID;
-        }))).on('close', callback);
+        spawn(
+            'kill',
+            ['-s', stripPrefix(signal), child.pid].concat(
+                kids.map(
+                    function(p) {
+                      return p.PID;
+                    }
+                )
+            )
+        ).on('close', callback);
       });
     } else {
       exec('kill -s ' + stripPrefix(signal) + ' ' + child.pid, function () {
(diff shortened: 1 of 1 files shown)


knight:
@remy is this PR any good?

remy:
Super late to this PR, but can you let me know why this would be needed? Is it causing breaking elsewhere? (genuine question).

knight:
On Linux auto re-spawning after detected changes does not work because sh does not recognize the SIGNAL constants with prefix. Please see my comments on the issue #956

heisian:
@knight why not do it the way an answer in the link you referenced suggests?
```
#!/bin/sh

set -x
set +o posix
sleep 100 &
pid=$!
kill -s SIGSTOP "$pid"
sleep 2
kill -s SIGCONT "$pid"
set -o posix
```

remy:
Fixed (alternative method) in nodemon@1.13.2

PR #962, opened 2017-01-05
Fix executable path handling under windows

attilah:
- If the user configured a *nix compatible npm script where the path contained a forward slash, that failed under windows, since cmd.exe treated that as a command line argument.
- The quote handling was not 100% correct under windows, the whole path needs to be quoted, not just the segments it is also modified.

### DIFF
--- a/lib/monitor/run.js
+++ b/lib/monitor/run.js
@@ -12,6 +12,7 @@ var noop = function () {};
 var restart = null;
 var psTree = require('ps-tree');
 var hasPS = true;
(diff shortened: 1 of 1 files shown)


attilah:
@remy,it is really "fun" that CI is failing, so is the PR. Hence other PRs are showing failure too :-(

ifeltsweet:
Can you please merge this?

Our team is working both on windows and macos and npm scripts are now broken as described above.

PR #985, opened 2017-02-19
Update travis ci environment

### LINKED ISSUE #981: Update node versions, in travis ci environment
Missing `7` version and still using `0.10`, need to consider dropping since it's no longer supported in mainstream. Latest stats that I could find, shown in here – https://semaphoreci.com/blog/2016/10/12/nodejs-versions-used-in-commercial-projects-2016-edition.html

~ 20% of usage, should be much less now

markelog:
Fixes #981


remy:
Going to park this PR, I think the upcoming PRs that I've merged should solve this (along with getting node 8 in there).

Issue #988, opened 2017-02-19
Mitigate `MaxListenersExceededWarning` error in tests


markelog:
Observing this at the test run
>  require-able
    ✓ should prioritise options over package.start (140ms)
    ✓ should know nodemon has been required
(node:27038) MaxListenersExceededWarning: Possible EventEmitter memory leak detected. 11 data listeners added. Use emitter.setMaxListeners() to increase limit
    ✓ should restart on file change (1228ms)
    ✓ should be restartable (1100ms)

Need to figure out why and what is happening and fix it

remy:
Heh, yeah, I couldn't work this out either!

remy:
I'd recommend using the latest nodemon, but if this is happening upon load,
then it's likely to be your own code.

The tests throw this error when nodemon is required an restarts *a lot*
during the tests.

Issue #983, opened 2017-02-19
Consider switching from jscs/jshint to eslint


markelog:
If accepted, we need to decide on eslint config

remy:
I'm open to this, but it needs to be in an atomic PR/commit. I'm just wary it might touch a lot of files (which is fine) but I just want to keep the merge limited to linting changes.

markelog:
It definitely will, no fear though, I already had experience with implementing such configs into big/medium code-bases in one commit, excluding proprietary systems, I can name for example grunt, jquery, wikimedia.

We only need to settle on what kind of code-style do you prefer :)

PR #990, opened 2017-02-19
WIP: working with tests stability

markelog:
Includes commits from some other open pulls, I think you can it's obvious where I am going with this


markelog:
Look at that, travis is green, still need to implement proper server clean-up, remove warnings in latest node, clean git history, rebase with master.

But before next iteration it would be cool to have a feedback in case i'm not going in the right direction

remy:
What was the change that caused them all to start passing? I'm always wary
that I cant see all the tests (there's about 100 instead of 200…but I could
be wrong, viewing on my mobile ATM).

Also, can you follow the commit message format for all commits so that the
release and notes get generated properly (can't remember if the format is
specified anywhere - again, on mobile).

markelog:
> What was the change that caused them all to start passing?

Some of them weren't cleaned properly i.e. in `After`/`AfterEach` some of them still are, those are currently skipped and I also started to divide them on `unit` and `integration` tests, that also helps. 

By no way I'm finished with this, but before I go further with this, it would be cool if you would give me a green light on continuing, so I would know if i'm going in the right direction, so this approach would adhere to you preferences.

> Also, can you follow the commit message format for all commits so that the
release and notes get generated properly (can't remember if the format is
specified anywhere - again, on mobile).

After i'm finish with it, I'm gonna squash, format, rebase and stuff, right now it's a WIP

remy:
Okay, everything in this PR looks legit to me 👍 good work. 

Happy for it to be squash/merged too (though I notice that the signal config has slipped in as a feature inside of the test tidy up, I don't mind, but if that's in a single commit, that would be great).

I assume these skipped tests are being tracked and can be restored once the test tidy up is further along (though the fact that it passed first time is a big step forward).

I also see you've managed to ditch my hack to run each test file individually - good work (I've long moved over to tap for testing, but I think it would be more work than it would be worth to convert all these tests away from mocha…just a side note!)

remy:
Ping me on this thread when you're ready to merge (and I can quickly review).

markelog:
Sounds good, meanwhile status update:
there is no more skipped tests anymore, had to remove one flacky one though, since it was unstable on systems with limited resources.

As a bonus there is no more ghost test files anymore after tests are executed and there is no more global mocha timeout, plus test clean-up now should be more stable.

There is still four issues that I would like to finish working on here:
1. Right now ghost test files will be removed if test run is not killed forcefully - i.e. if there is no terminate signals during test execution
2. In some situations (TBD) ghost httpd servers are left after test execution 
3. Need to do general cleaning squash and stuff
4. Would like to address https://github.com/remy/nodemon/issues/988 here also

Probably gonna do this in the next weekend, hopefully. Will ping you up as soon as I'm done. 

As a side note: I noticed sometimes code-style is inconsistent, i.e. https://github.com/remy/nodemon/issues/983 issue, I have couple ideas what code-style we can use and I probably can implement it

markelog:
Working on 1,2 items (mentioned above), still need to finish up: generalize & simplify the solution. Will try to work on this during the weekdays also, to speed things up

remy:
👍

remy:
First of all: ❤️ for all your hard work. 

I've recently pushed a load of changes to deps and updated the base level support for nodemon to match LTS node releases (which helps tests a lot since things like 0.10 are dropped).

I'm going to close this PR as the last series of test runs have all passed (when they should have) rather than breaking due to random race conditions (though…I'm not 100% certain they're all gone, but it's much better now).

Thank you again.

markelog:
No worries! And sorry I couldn't finish it, was feeling guilty about this :/

PR #1021, opened 2017-05-05
Better readability

kud:


remy:
I'll try to pull these changes in via another commit (I noticed there's a stray backtick in your version too though).

PR #1022, opened 2017-05-05
Better readability

kud:


remy:
Given you're just changing my 4 spaces for code to fenced can you put these PRs together​, commit them as (something like) `docs: spaces to fences` and use `[skip ci]` in the body (it'll skip the ci tests).

kud:
I was mostly interested here about syntax highlighting. And I did it via github website, so it was hard to modify two files at the same time. But yes, I will do it. ;)

Issue #1008, opened 2017-04-04
Ignoring all node_modules except for one


shawninder:
I'm trying to have nodemon ignore all node_modules except for one (a dependency I am developping concurrently with my main project), but I can't find any way to do this. Glob patterns would support this use-case, and so would regular expressions. However, since these ignore strings are converted from a glob-style pattern to regular expressions before being passed to chokidar, I can no longer find any way to include negations in my ignore strings.

For example:

**nodemon.json**

```json
{
  "ignoreRoot": [
    ".git",
    ".nyc_output",
    ".sass-cache",
    "bower-components",
    "coverage"
  ],
  "ignore": [
    "node_modules/!(is-node)",
    "node_modules/is-node/node_modules"
  ]
}
```

Here the negation character `!` simply becomes a litteral `!` in the regular expression sent to chokidar.
A similar thing happens when I try negative look-ahead syntax:

```json
{
  "ignoreRoot": [
    ".git",
    ".nyc_output",
    ".sass-cache",
    "bower-components",
    "coverage"
  ],
  "ignore": [
    "node_modules/(?!is-node)",
    "node_modules/is-node/node_modules"
  ]
}
```

@remy Is there anything in the existing stable nodemon which could support this use-case of ignoring all node_modules except for one of them?

**more details**: I made a stackoverflow question for this with more details before I had understood why this wasn't working.

remy:
Honestly, I'm not sure. I wonder if a PR that passes the pattern directly to chokidar would fix it? Though I'm really not sure of the impact of such a change…

shawninder:
I'm pretty sure it would fix it, but it might indeed be tricky to allow this without breaking things in existing code. Nodemon would have to differentiate the current pattern from a regular expression, but the path separator `/` and regular expression literal character `/` are the same character, so this is not trivial.

If, in addition to `nodemon.json`, nodemon also considered a `nodemon.js`, then we could then pass in a real regexp object instead of just strings... Just a thought

shawninder:
@remy I made a PR which solves the issue: https://github.com/remy/nodemon/pull/1027

Care to take a look?

remy:
Thank you for the work, and I hope this doesn't discourage, but I can't merge a change that makes nodemon even more complicated. Adding *another* way to configure nodemon just makes it harder to understand how to use it. I'd merge a change that silently fixes this, but I can't by adding nodemon.js

shawninder:
[...]

PR #1027, opened 2017-05-11
Support `nodemon.js` in addition to `nodemon.json`

shawninder:
Having the option of a `js` file in addition to a json file allows using real Regular Expression objects rather than strings, among other things.

This PR would solve https://github.com/remy/nodemon/issues/1008


remy:
Closing as per comment on issue. Sorry, I know this sucks.

PR #1028, opened 2017-05-13
Support configuration from 'nodemonConfig' in package.json

rhodgkins:
Allow `nodemonConfig` to be used in a projects `package.json`.
This matches the option of a lot of other development tools - [ESLint as `eslintConfig `](http://eslint.org/docs/user-guide/configuring#configuration-file-formats), [JSHint as `jshintConfig`](http://jshint.com/docs#Configuration), [JSCS as `jscsConfig`](http://jscs.info/overview#-config-c) but to name a few...!

This feature is entirely additive - if the `--config <file>` is specified or there's a local `nodemon.json` then `nodemonConfig` in the `package.json` will be ignored.


emilio-martinez:
I personally never put the config for nodemon or any of the packages mentioned above in package.json. I always choose to set config on a separate file.

mattstyles:
+1 always prefer config in one place and as package.json is a load of config it makes sense. Helps keep config relatively small as well.

tdd:
I try and centralize all my configs (Babel, ESLint, StandardJS, Browserlist, etc.) in a single package.json file for my apps, so this would be neat.

Issue #1031, opened 2017-05-18
nodemon can't read config files encoded with BOM e.g. UTF-8-BOM


aresilek:
If you create a nodemon.json configuration file encoded with BOM such as UTF-8-BOM nodemon fails to load the configuration with the error:
SyntaxError: Unexpected token ? in JSON at position 0
    at Object.parse (native)
    at <%APPDATA%>\npm\node_modules\nodemon\lib\config\load.
js:180:23
    at tryToString (fs.js:456:3)
    at FSReqWrap.readFileAfterClose [as oncomplete] (fs.js:443:12)
[nodemon] Failed to parse config <Project folder>\nodemon.json

The issue seems to be that fs doesn't strip BOM characters by design and JSON.parse [correctly] chokes when it sees the BOM character as invalid JSON.

See http://stackoverflow.com/questions/24356713/node-js-readfile-error-with-utf8-encoded-file-on-windows

PR #1032, opened 2017-05-18
Fix: nodemon can't read config files encoded with BOM e.g. UTF-8-BOM #1031

aresilek:
Fix: nodemon can't read config files encoded with BOM e.g. UTF-8-BOM #1031


remy:
I've got a feeling this was merged manually and was included in the 1.12.0 release.

aresilek:
Hi Remi, you committed this here: 0e4b310c1c27b5783a1d61f373a95110e21d8ee8

PR #1053, opened 2017-06-21
fix: catch module.parent.filename === undefined

mikeifomin:
It happened with electron.js. Maybe some wrapper module has a native nature


lib/utils/index.js
```diff
@@ -16,7 +16,11 @@ var utils = module.exports = {
   isLinux: process.platform === 'linux',
   isRequired: (function () {
     var p = module.parent;
-    while (p) {
+    while (p) { 
+      // in electron.js engine it happens
```
remy: yoinks!

PR #1074, opened 2017-07-28
Option to prevent checking of updates

uzyn:
```
nodemon --no-update-check
```

to prevent `update-notifier` from executing and calling home.


remy:
I think a bump of update-notifier should be the only change required, since update-notifier looks for these values automatically: https://github.com/yeoman/update-notifier#user-settings

PR #1077, opened 2017-07-28
Add envvar (NODEMON_PROCESS_STAGE) indicating first run or subsequent runs

### LINKED ISSUE #1075: Send an extra environment variable on restart only
I'm running Nodemon to live reload my Express app during development.

One of the things I do is I open a browser window with the root page when the server is started. But I've had to disable that, because when updating any code, Nodemon causes a restart, which causes the page to be reopened.

If I could get an extra environment variable that is only set on restart, I could just skip the code that opens the browser window.

msikma:
This adds a `NODEMON_PROCESS_STAGE` environment variable to the spawned process that is 1 during the initial run, and 2 during subsequent runs. This allows the user to skip certain initialization routines that are not necessary on restart.

Fixes #1075. Given that the project is inactive I doubt this will get merged, but it's here for posterity and in case someone else needs this. I'll probably publish my fork on NPM under another name.

### DIFF
--- a/faq.md
+++ b/faq.md
@@ -93,6 +93,12 @@ Say you did want to watch the `node_modules` directory. You have to override the
 
 Now when ignoring `public`, the ignore rule results in `['.git', 'public']`, and nodemon will restart on `node_modules` changes.
 
+## How do I check from inside my script that it's running on nodemon?
+
+nodemon injects a single additional environment variable into the process: `NODEMON_PROCESS_STAGE`. Its value is `'1'` if nodemon has not restarted your process yet, and becomes `'2'` after subsequent restarts.
+
+If you have any initialization that you would like to do only when running nodemon and only one time (e.g. printing some help information to the terminal), you can do so by checking if `NODEMON_PROCESS_STAGE` equals `'1'`.
+
 ## nodemon doesn't work with fedora
 
 Fedora is looking for `nodejs` rather than `node` which is the binary that nodemon kicks off.
--- a/lib/monitor/run.js
+++ b/lib/monitor/run.js
@@ -13,6 +13,12 @@ var restart = null;
 var psTree = require('ps-tree');
 var hasPS = true;
 
+// Whether the process has been started once before.
+// 0 = initial state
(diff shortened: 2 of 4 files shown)


remy:
I'm open to merging this feature, but can we add docs and tests to the PR too?

msikma:
Thanks for having a look! Yeah, I will need to find some time to write those. Maybe this weekend I will look into this...it's busy.

msikma:
I've added [an entry to the FAQ](https://github.com/msikma/nodemon/blob/d8cbb84ed62b9ea9f216a4137386d3c43d94ace4/faq.md#how-do-i-check-from-inside-my-script-that-its-running-on-nodemon) and made a unit test that checks if the value correctly becomes `'1'` and then `'2'` when restarting. I'm not sure just how good the test is, though...maybe you could have a look and tell me if it meets with your standards.

remy:
I'm going to close this. I'm not sure the implementation is quite right, and I wonder if it's more useful to expose the current nodemon state (see [events.md](https://github.com/remy/nodemon/blob/master/doc/events.md#states)) via an environment value.

There's also whether this is actually an edge case, or whether there's real demand for this value. I'm open to hearing from others via this PR (if you want to bring them in).

Thank you all the same.

msikma:
It is probably an edge case, because I didn't see anyone else requesting it, but I wrote this because it just seems like good design to be able to tell whether your program is running in reload mode, and whether it is in the first or nth iteration of that cycle. The simplest use case is not displaying a "welcome" message on subsequent reloads. But you can also use it to skip expensive checks that make sense the first time but not if you're constantly reloading the software, or skip child processes from being spawned, etc.

It's fine if you don't want to merge this, but I'd rather have heard that before I spent time writing tests for a feature you don't want...

remy:
Sure, and I'm sorry about that. I appreciate the hard work that goes into just making a PR in the first place. I do insist that all PRs come with tests (pretty sure that's documented), but I didn't run through the impact of the change properly, and that's my bad.

Issue #1095, opened 2017-09-13
update to v1.12.0 cause crash when pass argument to node


lerit:
i update nodemon from v1.11.0 to v.12.0
but it fail when i use it.
this is my nodemon.json:
  "execMap": {
    "js": "node --harmony --trace-warnings"
  },
the error mean:"node --harmony --trace-warnings" Is Not internal or external commands
i must remove the argument or degrade to version 1.11.0, and it work fine again
![1 11](https://user-images.githubusercontent.com/4743467/30356033-82ec9132-9869-11e7-84dd-33a015ec678a.png)
![1 12](https://user-images.githubusercontent.com/4743467/30356034-82ef055c-9869-11e7-9e8b-bca77e4efd28.png)

lpgera:
I am getting the same error with node --trace-warnings, after updating to nodemon 1.12.0.

Windows 64bit, Node 8.3.0

```
[nodemon] starting `node --trace-warnings ./src/server.js`
[nodemon] child pid: 28200
'\"node --trace-warnings\"' is not recognized as an internal or external command,
operable program or batch file.
[nodemon] app crashed - waiting for file changes before starting...
```

remy:
Can someone try a PR to fix this windows issue? Obviously ppl are seeing
it, i just don't have Windows to hand to fix as easily. Thanks in advance.

Issue #885, opened 2016-07-28
Cannot clone and checkout branch due to fixture filenames on Windows


AlmirKadric:
This project is not usable on `Windows` as some of the test fixtures have filenames with unsupported characters. This causes `Git` to break during it's checkout phase.

An example fixture filename is `some\"file`

pombredanne:
and this and the dupe #900 make scancode choke https://github.com/nexB/scancode-toolkit/issues/413 when scanning nodemon

PR #1096, opened 2017-09-14
Fix: crash when passing argument to node #1095

lpgera:
This seems to fix issue #1095 but I could not run tests because of the issue with checking out the repository on Windows: #885


remy:
You absolute star. nodemon 1.12.1 should be (automatically) released shortly.

PR #1102, opened 2017-09-23
Allow script be configured through nodemon config file

birdbrained:
Allow script be configured through `nodemon.json`

### DIFF
--- a/lib/config/load.js
+++ b/lib/config/load.js
@@ -30,7 +30,10 @@ function load(settings, options, config, callback) {
       // Then merge over with the user settings (parsed from the cli).
       // Note that merge protects and favours existing values over new values,
       // and thus command line arguments get priority
+      
+      let script = options.script;
       options = utils.merge(settings, options);
+      if (!options.script) options.script = script;
 
       // legacy support
       if (!Array.isArray(options.ignore)) {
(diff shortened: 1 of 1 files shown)


lib/config/load.js
```diff
@@ -30,7 +30,10 @@ function load(settings, options, config, callback) {
       // Then merge over with the user settings (parsed from the cli).
       // Note that merge protects and favours existing values over new values,
       // and thus command line arguments get priority
+      
+      let script = options.script;
       options = utils.merge(settings, options);
+      if (!options.script) options.script = script;
```
remy: I appreciate this a very old PR, but I'm not sure how this isn't happening already? If `options.script` exists, it'll land in the `options` value.

remy:
I'm closing this PR based on [this](https://github.com/remy/nodemon/pull/1102/files#r154668429) - but happy to reopen if this is indeed a bug!

lib/config/load.js
birdbrained: I remember finding this causing an issue where I couldn't get something working because options were getting overwritten somewhere.

birdbrained: This is old, it may not apply anymore if someone else fixed it. I'll try and remember exactly what was happening.

birdbrained: After trying to find the cause of my issue here again: I remember debugging this and finding that options.script was not defined even though it was in the config. I can't find the project I was working on that had this problem, I think I have a monkeypatchched version of nodemon somewhere though to work around it.

birdbrained: maybe settings has a script property that evaluates to false?

PR #1110, opened 2017-10-04
fix: read config file before defaulting script parameter

lennym:
If I have a nodemon.json config file with an `exec` property, then this is automatically appended with ` index.js` when running `nodemon` standalone.

I would expect this to be equivalent to running `nodemon --exec "my cmd"` - that is providing an `exec` prevents defaulting of `script` to index.js.

To resolve, move the code which defaults the `script` property when `exec` is undefined to after the config files have been read, so that the behaviour is the same irrespective of whether properties are set in CLI flags or in nodemon.json.

### DIFF
--- a/lib/cli/parse.js
+++ b/lib/cli/parse.js
@@ -85,17 +85,6 @@ function parse(argv) {
     }
   }
 
-  if (script === null && !nodemonOptions.exec) {
-    var found = findAppScript();
-    if (found !== null) {
-      if (found.exec) {
-        nodemonOptions.exec = found.exec;
-      }
-      script = found.script;
-      nodemonOptions.scriptPosition = args.length;
-    }
-  }
-
   nodemonOptions.script = script;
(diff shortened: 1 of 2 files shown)


lennym:
As an unrelated aside, I wrote a first version of the commit message for 8a4dc777ee5f3fd12123f23f5bc571d639f77872 and tried to commit before I realised that you have `commitlint` enabled, which then discarded my lovingly crafted message because I didn't have `fix: ` as the first line.

I've no problem with `commitlint` as a concept, but it'd be good to get some warning as a contributor before it discards my messages.

Where am I going with this..?

I wrote a module in response to this slight that allows for custom per-project commit templates - https://www.npmjs.com/package/commit-template. I might suggest its use for output something like:

> \# Warning: this project uses commitlint. Your message must be in the following form or it will not be accepted:
> \# &lt;example message&gt;

PR #1119, opened 2017-10-19
fix: Clarified that "mounted drives" refers to mounted network drives in Readme.

dabockster:
Changed the "Application isn't restarting" to clarify that mounted drives are, in fact, mounted network drives in this case (since drives that are mounted do not necessarily have to be via the network).

### DIFF
--- a/README.md
+++ b/README.md
@@ -179,7 +179,7 @@ Note that by default, nodemon will ignore the `.git`, `node_modules`, `bower_com
 
 ## Application isn't restarting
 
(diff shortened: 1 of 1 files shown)


mvcds:
It's not clear for me what a mounted drive would be in this case.

A drive which hasn't been initialized yet? Can you point me to some examples, pls?

remy:
I'm not sure this is accurate change to be honest. This flag referrers to mounted drives both over network, but more commonly between a virtual machine/instance and the host machine.

I'm going to close (with thanks), as I think it doesn't quite work.

PR #1134, opened 2017-11-17
Use child_process.fork when running node scripts...

heisian:
* Handles cleanup of node child processes gracefully.
* Doesn't pass tests, but I'm too lazy to go into this any further.
* Regardless, `child_process.fork` is the correct way to handle node child processes.

### DIFF
--- a/lib/monitor/run.js
+++ b/lib/monitor/run.js
@@ -2,8 +2,9 @@ var debug = require('debug')('nodemon');
 var utils = require('../utils');
 var bus = utils.bus;
 var childProcess = require('child_process');
-var spawn = childProcess.spawn;
 var exec = childProcess.exec;
+var fork = childProcess.fork;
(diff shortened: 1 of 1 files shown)


remy:
Thank you for this PR, but I'm 99% certain that I'm not using fork due to the way that nodemon needs to communicate (via signals and streams) with the child process. I think you can see this in the failing tests (i.e. they don't work because the code has changed to use fork).

I'm going to close for now. If you still this this is the right way for nodemon to execute the child, if you can get the tests to pass then it will help getting the change through.

heisian:
Sure, I will look into it. Fork can communicate over streams just as Spawn can, except fork defaults to IPC which is better for comms between node processes, IMO. So you or I would either need to make the fork communicate via streams, or add a whole set of conditions for when to use IPC and when to use stdio/out. Should have some time in the coming weeks b/c of the holidays.

Issue #690, opened 2015-10-18
Fails to watch directory with exclamation mark in name (windows)


jdoose:
nodemon fails to watch the files in a directory, if the directory name contains an exclamation mark. 
This happens under Windows, cannot confirm for other systems.
When the directory of the node js file contains a quotiation mark it reads " [nodemon] watching 0 files". If I rename the directory to the same name _without_ exclamation mark it works.

Using pure chokidar it works in both cases.

I guess it has something to do with parsing ignore/include patterns, but that is just a rough guess.

Nodemon version 1.7.1
It worked when using nodemon 1.3.x

Dump output

```
C:\doose-Data\Work\Raup_Ritter\CHECK!Lehrerzimmer\sub1>nodemon --dump --verbose --debug app.js
  nodemon bus new listener: reset (0) +0ms
  nodemon bus new listener: reset (0) +5ms
  nodemon bus new listener: quit (0) +16ms
  nodemon bus new listener: quit (0) +1ms
  nodemon bus new listener: restart (0) +2ms
  nodemon bus new listener: restart (0) +2ms
  nodemon bus new listener: reset (2) +12ms
  nodemon bus emit: reset +2ms
  nodemon:watch resetting watchers +1ms
  nodemon reset +2ms
  nodemon config: dirs +236ms [ 'C:\\doose-Data\\Work\\Raup_Ritter\\CHECK!Lehrerzimmer\\sub1' ]
18 Oct 11:42:59 - [nodemon] 1.7.1
18 Oct 11:42:59 - [nodemon] to restart at any time, enter `rs`
18 Oct 11:42:59 - [nodemon] ignoring: .git C:\doose-Data\Work\Raup_Ritter\CHECK!Lehrerzimmer\sub1\node_modules/**/* bower_components .sass-cache
18 Oct 11:42:59 - [nodemon] watching: *.*
18 Oct 11:42:59 - [nodemon] watching extensions: js
--------------
node: v0.12.6
nodemon: 1.7.1
command: node C:\Users\doose\AppData\Roaming\npm\node_modules\nodemon\bin\nodemon.js --dump --verbose --debug app.js
cwd: C:\doose-Data\Work\Raup_Ritter\CHECK!Lehrerzimmer\sub1
OS: win32 x64
--------------
{ run: false,
  system: { cwd: 'C:\\doose-Data\\Work\\Raup_Ritter\\CHECK!Lehrerzimmer\\sub1' },
  required: false,
[...]

PR #1137, opened 2017-12-02
fix: Directories with ! mark + Failing tests #690

jarpi:
Exception not catched
Add tests for special characters path


remy:
@jarpi just an FYI - this option is now turned on by default in Windows, and can be set via config using `watchOptions: { disableGlobbing: true }` in `nodemon.json` (I need to document this too!)

jarpi:
@remy Already know, that's why I closed this :) 
Glad to see is partially fixed, even getting no credit out of it

PR #1142, opened 2017-12-05
fix: pass through execArgs from config

### LINKED ISSUE #1045: Make a property in config to pass additional arguments before `exec`
Please make a property (`args` or `runtimeArgs` for example) in `nodemon.json` config file to pass additional arguments before `exec`. It may be unknown for `nodemon` but valid for `node`, e.g. `--debug`, `--debug-brk`, `--inspect`, `--inspect-brk`, etc.

remy:
Also updating some deps and linting tweaks.

Fixes #1045

### DIFF
--- a/package.json
+++ b/package.json
@@ -36,7 +36,8 @@
     "postspec": "npm run clean",
     "clean": "rm -rf test/fixtures/test*.js",
     "web": "node web",
-    "semantic-release": "semantic-release pre && npm publish && semantic-release post"
+    "semantic-release": "semantic-release pre && npm publish && semantic-release post",
+    "postinstall": "node -e \"console.log('\\u001b[32mLove nodemon? You can now support the project via the open collective:\\u001b[22m\\u001b[39m\\n > \\u001b[96m\\u001b[1mhttps://opencollective.com/nodemon/donate\\u001b[0m\\n')\""
   },
   "devDependencies": {
     "@commitlint/cli": "^3.1.3",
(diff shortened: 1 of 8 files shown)


package.json
```diff
@@ -36,7 +36,8 @@
     "postspec": "npm run clean",
     "clean": "rm -rf test/fixtures/test*.js",
     "web": "node web",
-    "semantic-release": "semantic-release pre && npm publish && semantic-release post"
+    "semantic-release": "semantic-release pre && npm publish && semantic-release post",
+    "postinstall": "node -e \"console.log('\\u001b[32mLove nodemon? You can now support the project via the open collective:\\u001b[22m\\u001b[39m\\n > \\u001b[96m\\u001b[1mhttps://opencollective.com/nodemon/donate\\u001b[0m\\n')\""
```
zckrs: I understand your needs but this is really clumsy...

Imagines if all npm modules do that. :-/

Can you deactivate this log when we are in CI env?

remy: If all npm modules did this, then I'd find another way to ask and plea for help.

I've found myself asking this a lot lately, but: why are you using nodemon in CI?

zckrs: Just installing packages. 

Can you considers to move this console.log when nodemon is running for the first time ?

remy: As an arbitrary aim, if the project can raise $10-20 per week, then I'll happily remove the plea for support (note this wouldn't cover the past costs or even my ongoing costs, but it shows there's support that values nodemon).

Issue #1144, opened 2017-12-05
version 1.12.3 raised the minimum node version to 6.  This breaks our builds


Noyabronok:
version 1.12.3 raised the node version to >= 6 
https://github.com/remy/nodemon/compare/v1.11.0...v1.12.3

This is breaking installs that specify `engine-strict = true` in .npmrc and the node engine version listed in package.json is < 6.

Is 6 really the minimum supported version for nodemon, or can we set it to something less, like 4?

remy:
1.12.x is making use of more es6 which 4 may not have support for. 

Also: I'm pretty sure 4 is not supported and not part of long term support on node's roadmap (known and open security issues), but also: nodemon is a development tool, so why is it failing a build process? (Obviously I don't know the context of your software).

I can't support unsupported versions of node though.

Noyabronok:
Going from Node 0.8 to 6 should not be a minor but a major version increase.  Our local install is failing because we have `engine-strict=true` in our `.npmrc` file.   Node 4 supports a decent amount of ES6, and you can check the features here http://node.green/

We've been running nodemon under Node 4 without any problems for a while now.  Node 4 is officially supported as maintenance release until April 2018.  Can you please re-open this issue.

remy:
I'm inclined to agree that nodemon should have had a major bump. If you raise a PR that changes the engine down and adds Travis tests back in, and the tests pass, then I'll merge and release. How does that sound?

Noyabronok:
I created two pull request, choose one:
Revert down to Node 0.10: https://github.com/remy/nodemon/pull/1148
Revert down to Node 4: https://github.com/remy/nodemon/pull/1149

remy:
Just had a quick look at both and it looks like neither are passing tests.

Noyabronok:
The tests should be passing.  I'm seeing Travis fail during `npm install`
```
$ npm install 
Error: Failed to replace env in config: ${NPM_TOKEN}
```

remy:
@Noyabronok merge the latest commit from master and you'll get a fix in the `.travis.yml` file for that.

Noyabronok:
@remy Are you referring to https://github.com/remy/nodemon/commit/57c14a2b5867905247442e657f0942cebab35c6d ?
It's already included in my branch

remy:
Damn, yeah, I was. It is possible the `.npmrc` file is left in a cache on the test? (I'll take a quick look at both PRs and clean any caches).

[...]

PR #1148, opened 2017-12-06
revert ES6 usage and node version down to 0.10

Noyabronok:
Reverts ES6 changes and removes husky.  Tests pass under Node 4, but fail under 0.10.  At least it installs without errors under node 0.10.


remy:
Don't remove husky (and don't add your hidden project files). Husky is used to ensure commit messages style which allows for releases.

Noyabronok:
I had to remove husky for `npm install` to work under Node 0.10
I removed the hidden project files from PR

heisian:
wait why are you using Node 0.10?

Noyabronok:
@heisian version was originally 0.8 but I took the liberty to specify and test under 0.10.  Even with 0.10 tests are failing, but at least `npm install` succeeds.  There is [another PR](https://github.com/remy/nodemon/pull/1149) with Node 4 that has test passing

heisian:
yeah, but... why? are you restricted to using a version that was release 3 years ago and has since been superseded by not only a stable non-alpha/beta release (1.0) and on top of that **8 other versions**??

Noyabronok:
@heisian Please see https://github.com/remy/nodemon/issues/1144 for reasons for these PRs

heisian:
I read that PR, you're listing v0.10 and v4, and trying to implement both, when clearly v4 is really the only sensible option, if at all...

Noyabronok:
@heisian  I was asking @remy to pick the PR he prefers.  Node v4 works for me, and there's no reason that I see that it shouldn't. If nodemon doesn't work at all on 0.10 then we don't need that PR.

In general, the major node version in package.json should not change for a minor nodemon update.  A major version bump is warranted.  What I think should happen:

1) specify node v4 on master branch.  Bump up nodemon version to 2.0
2) if nodemon works on 0.10 - specify node 0.10 on a maintenance branch.  Bump current nodemon version at 1.13.x.

heisian:
@Noyabronok seems like a lot of trouble. i would just fork in this situation, or really just upgrade my node. but to each their own!

Noyabronok:
@heisian you gotta remember that not everyone has the luxury of updating their node version.  That's why it's important to support the node version specified in package.json, and to change nodemon's major version when removing support for older node version

remy:
Node 0.10 failing. Do you want to try to fix these or just close the PR and stick with node 4? @Noyabronok: Your call.

Noyabronok:
@remy Yeah, this is what I was talking about.  I made enough changes for install to work under 0.10 but tests fail.  If someone has a need for 0.10 or 0.8 they can pick up from here, so we can close this PR for now.  Let's stick win Node 4 for simplicity, since that's all I need right now.

PR #1149, opened 2017-12-06
node 4 min

Noyabronok:
enforce node 4 instead of node 6 as minimum.
all tests pass under node 4

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -18,3 +18,5 @@ node_modules
 coverage
 tmp
 test/fixtures/test*
+
+.idea
--- a/.travis.yml
+++ b/.travis.yml
@@ -9,7 +9,7 @@ node_js:
   - '8'
   - '6'
 before_install:
-  - if [ "$TRAVIS_BRANCH" == "master" ]; then echo "//registry.npmjs.org/:_authToken=\${NPM_TOKEN}" > .npmrc; fi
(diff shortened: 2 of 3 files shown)


remy:
Looks good. I'll squash the commits so they follow the right syntax for the commit style (i.e. to allow it to deploy)

Noyabronok:
@remy let's get this PR merged. Thanks!

remy:
Can you merge [this commit](https://github.com/remy/nodemon/commit/8967c02301cbc4f1f0cecab2e192e0ede82463cf) into your branch? It should fix the npm install problem properly now and will allow me to merge without having to edit your code.

remy:
Awesome work.

remy:
Kinda my issue with node 4: https://travis-ci.org/remy/nodemon/builds/315493550 😢

remy:
I know what this is now. Another PR requires node>4 :(

Noyabronok:
what pr is that?  Can it be updated to be compatible with Node 4?

remy:
If you check master out you'll see it failing on the config/env.js test in node 4.

It seems like run.js using fork in node 4 is throwing an exception when it tries to attach `child.stdout.on` but I don't see it in the test so it eventually times out.

remy:
Sorted and live.

PR #1156, opened 2017-12-12
Updated faq.md

eternalsayed:
Added: Solution for `Error: Cannot find module 'internal/util/types'` with ionic framework after updating node


remy:
Pushed a non-ionic-specific change to faq instead.

Issue #743, opened 2016-01-02
Can't watch all files in a directory


daveroberts:
Nodemon can't watch all files for change in a directory.  It looks like others are trying to pass --ext '.' for this, but that throws an exception.

```
/apps/home/daver/code/npm-scripts-example/node_modules/nodemon/lib/config/exec.js:158
  extension = extension.match(/\w+/g).join(',');
                                     ^

TypeError: Cannot read property 'join' of null
    at exec (/apps/home/daver/code/npm-scripts-example/node_modules/nodemon/lib/config/exec.js:158:38)
    at /apps/home/daver/code/npm-scripts-example/node_modules/nodemon/lib/config/load.js:56:29
    at callback (/apps/home/daver/code/npm-scripts-example/node_modules/nodemon/lib/config/load.js:161:5)
    at ReadFileContext.callback (/apps/home/daver/code/npm-scripts-example/node_modules/nodemon/lib/config/load.js:171:14)
    at FSReqWrap.readFileAfterOpen [as oncomplete] (fs.js:303:13)
```

nikdo:
I was also watching all files with nodemon, but I'm not sure if it is supported use case as it is not mentioned anywhere except some blogposts.

`--ext '.'` stopped working with #693.

If you still want it, the only workaround I have found is to stick with version 1.7.2.

musichen:
The same thing is here:
[nodemon] watching: *.*
[nodemon] starting `node --stack_size=4096 app.js`
/home/username/dev/bitbucket/myproject/node_modules/nodemon/lib/config/exec.js:158
  extension = extension.match(/\w+/g).join(',');
                                     ^
TypeError: Cannot read property 'join' of null

daveroberts:
This still looks to be the case.  I tried again with v1.12.5

dominykas:
Seems to be fixed in v1.13.3 - I ran with `--ext '.'`. nodemon restarted with updating `README.md` and `index.js`.

dominykas:
It still won't restart on extension-less files, e.g. `README` - not sure if that's something `nodemon` needs to support?

remy:
Shouldn't the "all" ext be `*` not `.` and then that matching rule updated for a special case? That said, it's straight forward to support both of these "special" rules in `match.js`

dominykas:
[...]

PR #1177, opened 2017-12-18
feat: support wildcard extension matching

dominykas:
Add support for `--ext ''`; add tests for `--ext '.'` and `--ext '*'`.

Caveat: does not match extension-less files, see: https://github.com/remy/nodemon/issues/743#issuecomment-352405647.


lib/config/exec.js
```diff
@@ -186,8 +190,8 @@ function exec(nodemonOptions, execMap) {
   options.ext = extension;
```
dominykas: Should I make it `undefined` of even `delete` it when it is an empty string?

PR #1178, opened 2017-12-18
chore: add .npmrc which explicitly enables package-lock; some minor doc tweaks

dominykas:
I also noticed that `jshint` gives a ton of errors when used with the `.jshintrc` in the repo - is still worth keeping around as it seems it is not used anyways?

### DIFF
--- a/.github/CONTRIBUTING.md
+++ b/.github/CONTRIBUTING.md
@@ -2,7 +2,7 @@
 
 ## Commit messages
 
-Commit messages must follow the [Angular-style](https://github.com/angular/angular.js/blob/master/CONTRIBUTING.md#commit-message-format) commit format (but excluding the scope).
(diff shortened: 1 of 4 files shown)


remy:
This npmrc file will be blown away during the Travis process. Is that expected?

dominykas:
Ah. Travis writes an auth-token in it... which means Travis will be using default npm settings (and this is the default), so I guess it does no harm? Do you allow Travis to push back to the repo for semantic release (I don't know how all of that stuff works)? If so - then it might remove this file... or even worse - commit the credentials it used into the repo?

dominykas:
I can update `.travis.yml` to use `>>` rather than `>`... npm ignores it by default when publishing, so as long as semantic-release does not push back to the repo - it should be fine?

remy:
No, Travis doesn't push back, only deploys to npm, but yeah, if it's going to ensure the package-lock contents then it makes sense for the Travis yml file to append.

PR #1179, opened 2017-12-18
test: fix hanging test

dominykas:
I can't explain this. Sorry :)

The test were hanging randomly and intermittently for me - so I started debugging and adding `console.log`s here and there - and all of a sudden they started passing. So I figured it's something to do with delaying the execution? Possibly something related to promises and event loops and watches created in the same tick. Unsure. But adding a timeout seems to make things a _lot_ more stable? 

I assume the last test was hanging for the same reason?

I could also be wrong... but worth trying?

### DIFF
--- a/test/config/env.test.js
+++ b/test/config/env.test.js
@@ -9,8 +9,7 @@ describe('when nodemon runs (1)', function () {
   after(function (done) {
     // clean up just in case.
     nodemon.once('exit', function () {
-      nodemon.reset();
-      done();
+      nodemon.reset(done);
     }).emit('quit');
   });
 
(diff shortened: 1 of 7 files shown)


remy:
I tend to run `killall node` before I kick off a test. I know there can be sometimes race conditions that make testing tricky (sometimes I'll just use `npx mocha <test-file>` to sign off before pushing to Travis)

dominykas:
Ah, yes, I did notice that sometimes it does not always kill all child processes (seems there's a stray `node test/fixtures/sigint.js` remaining), however I do get failures even on a clean slate - seems to be related to restarts happening more often than desired in the test. I've replaced `unlink` with `unlinkSync` and that _feels_ a little more stable, but not there just yet...

dominykas:
The final commit is where it's at (increasing the timeout... this is starting to feel like a regular day at work...), but I figure the remaining changes are also worth having?

* changed all `nodemon.reset()` calls in the tests to actually use wait for completion, i.e. passing in `done`
* changed all `unlink()` to be `unlinkSync()`
* calling unlink after _each_ test to keep clean slate
* checking file exists before attempting to unlink

They probably make the run somewhat slower, but at least I'm getting green runs locally...

PR #1201, opened 2017-12-29
Watch both js and mjs files

MajorBreakfast:
- Watch both `js` and `mjs` files if the script is either a `js` or `mjs` file or file with no extension. (For some reason that was before only the case when the main script was a CoffeeScript file xD)
- Some cleanup

### DIFF
--- a/lib/config/exec.js
+++ b/lib/config/exec.js
@@ -113,9 +113,12 @@ function exec(nodemonOptions, execMap) {
 
   var scriptExt = path.extname(script).slice(1);
 
-  var extension = options.ext !== undefined ?
-      options.ext :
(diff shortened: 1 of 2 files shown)


remy:
Hopefully you've seen it, but the tests need to be updated with that to expect on the extension (also thanks for PR'ing this, been meaning to open the issue myself 👍)

MajorBreakfast:
Ah... the test hadn't completed when I last looked. Fixing it now... ;)

MajorBreakfast:
Let's see if that did the trick.

remy:
Great stuff. Should be live once Travis finishes testing `master` 👍

MajorBreakfast:
Thanks for handling this so fast!

PR #1224, opened 2018-01-11
fix: don't ignore dot-directories

### LINKED ISSUE #1223: Nodemon won't restart app
After saving files, nodemon does not restart app, giving the message:
`[nodemon] changes after filters (before/after): 1/0`

Spending several hours I found the solution.
In Windows OS if your have dot in path of your project it displays the message above as soon the dot is removed everything works as expected.

`c:\.projects\your-app` change to `c:\project\your-app` - starts work 👌.

So is it bug or some limitation?

remy:
Fixes #1223

### DIFF
--- a/lib/monitor/match.js
+++ b/lib/monitor/match.js
@@ -1,7 +1,8 @@
-var minimatch = require('minimatch');
-var path = require('path');
-var fs = require('fs');
-var utils = require('../utils');
+const minimatch = require('minimatch');
+const path = require('path');
+const fs = require('fs');
+const debug = require('debug')('nodemon:match');
+const utils = require('../utils');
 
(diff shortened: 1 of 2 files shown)


lib/monitor/match.js
```diff
@@ -167,12 +168,16 @@ function match(files, monitor, ext) {
...
+  debug('rules', rules);
+
   var good = [];
   var whitelist = []; // files that we won't check against the extension
   var ignored = 0;
   var watched = 0;
   var usedRules = [];
-  var minimatchOpts = {};
+  var minimatchOpts = {
+    dot: true,
```
remy: This is the actual fix.

PR #1270, opened 2018-02-26
fix: make it possible for windows to checkout

### LINKED ISSUE #1255: Cannot checkout sources on Windows
Windows 10

### Expected behaviour

Nodemon sources can be checked out on Windows.

### Actual behaviour

Git checkout fails with error 'error: unable to create file test/fixtures/some\"file: No such file or directory'. One of test file names are not compatible with Windows

One possible way to address this would be not to store files in question in repo but create during test run on compatible platforms.

remy:
Fixes #1255

### DIFF
--- a/test/fork/run-mac-only.test.js
+++ b/test/fork/run-mac-only.test.js
@@ -12,11 +12,11 @@ const filenames = [
 if (!process.env.TRAVIS && process.platform !== 'win32') {
   describe('nodemon fork (mac only)', () => {
     before(() => {
-      filenames.map(([filename, contents]) => fs.writeFileSync(filename, contents, 'utf8'));
+      filenames.map(file => fs.writeFileSync(file[0], file[1], 'utf8'));
     });
 
     after(() => {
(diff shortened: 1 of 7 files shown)


test/fork/run-mac-only.test.js
```diff
@@ -0,0 +1,82 @@
...
+const utils = require('../utils');
+const appJS = utils.appjs;
+const run = utils.run;
+
+const filenames = [
+  [__dirname + 'some\\\"file', '#!/usr/bin/env node\nconsole.log("OK");'],
+  [__dirname + 'some\ \\file', '#!/bin/sh\necho "OK"'],
+];
+
+if (!process.env.TRAVIS && process.platform !== 'win32') {
```
igor-savin-ht: Why not run this on Travis?

remy: I don't remember, but it was carried across from the original tests. I suspect it made things blow up, and it was a very, very specific test.

Issue #1269, opened 2018-02-26
Restart is not triggered for files in a directory.


brkc:
- `nodemon -v`: 1.15.1
- `node -v`: v8.9.4
- Operating system/terminal environment: Windows 10 Pro x64 / cmd
- Command you ran: `nodemon -C --watch broken\touchme.txt --exec "echo changed"`

### Expected behaviour
When I change the file, it prints "changed".

### Actual behaviour
It does not print "changed".

---

This works in 1.14.12.

```
> npx nodemon@1.14.12 -C --watch broken\touchme.txt --exec "echo changed"
...
[nodemon] 1.14.12
[nodemon] to restart at any time, enter `rs`
[nodemon] watching: broken\touchme.txt
[nodemon] restarting due to changes...
[nodemon] starting `echo changed`
changed
[nodemon] clean exit - waiting for changes before restart
```

It does nothing in 1.15.1.

```
> npx nodemon@1.15.1 -C --watch broken\touchme.txt --exec "echo changed"
...
[nodemon] 1.15.1
[nodemon] to restart at any time, enter `rs`
[nodemon] watching: broken\touchme.txt
```

remy:
I think this might be a dupe issue. Don't have my laptop to hand to confirm though.

brkc:
In any case, here's some additional debug info in case it's helpful.

1.14.12

```
> cross-env DEBUG=nodemon:* npx nodemon@1.14.12 --verbose -C --watch broken\touchme.txt --exec "echo changed"
...
[nodemon] 1.14.12
[nodemon] to restart at any time, enter `rs`
[nodemon] or send SIGHUP to 14248 to restart
[nodemon] watching: broken\touchme.txt
[nodemon] watching extensions: js,mjs,json
[nodemon] spawning
[nodemon] child pid: 24048
  nodemon:watch watching dir: broken\touchme.txt +0ms
[nodemon] watching 1 file
[nodemon] files triggering change check: Broken\touchme.txt
  nodemon:watch filterAndRestart on [ 'Broken\\touchme.txt' ] +10s
  nodemon:match rules [ '!**\\**/.git/**',
  '!**\\**/.nyc_output/**',
  '!**\\**/.sass-cache/**',
  '!**\\**/bower_components/**',
  '!**\\**/coverage/**',
  '!**\\**/node_modules/**',
  '**\\broken\\touchme.txt' ] +0ms
  nodemon:match match C:\w\test\Broken\touchme.txt true +8ms
[...]

PR #1272, opened 2018-02-26
fix: put windows drive letter tweak in right place

remy:
Ref #1269

In progress. No merge.


remy:
Damnit - node 4 doesn't have destructuring. Will have to tweak some files.

PR #1285, opened 2018-03-06
Fix: Update README.md

patitonar:
I included html and css as examples on files to monitor. I had to make this work on a project and I had to get the answer of how to do this from stack overflow because it wasn't clear. I think this will help others with the same situation.

### DIFF
--- a/README.md
+++ b/README.md
@@ -177,7 +177,7 @@ By default, nodemon looks for files with the `.js`, `.mjs`, `.coffee`, `.litcoff
 nodemon -e js,jade,html,css
 ```
 
-Now nodemon will restart on any changes to files in the directory (or subdirectories) with the extensions .js, .jade.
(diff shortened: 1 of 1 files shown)


remy:
Why would you want or need to restart node (via nodemon) for html or CSS changes? They're not persistent in memory unless you've got a bespoke caching system in place.

patitonar:
I'm using express to server-side render a webpage loading a base html and css file when the server starts. Every time I change the html or css I had to restart node so it loads the new base html/css files. 
Using `nodemon -e js,html,css` saved me a lot of time.

remy:
When using express, you do not need to reload node when HTML or CSS changes. In fact, you don't need to reload node when your views change either.

You're actually wasting time on the reload process when making those changes.

My typical setup for express is:

```
$ nodemon --ignore public --ignore views
```

So it only watches `.js` and ignores the static content. 

I'm closing this PR as you shouldn't be reloading node at all for static content. Thank you all the same though.

PR #1313, opened 2018-04-13
cs: refactoring variables

RobinCK:
Code consistency
Sometimes var was used and sometimes const

### DIFF
--- a/lib/cli/index.js
+++ b/lib/cli/index.js
@@ -1,3 +1,4 @@
+'use strict';
 const parse = require('./parse');
 
 /**
--- a/lib/cli/index.js
+++ b/lib/cli/index.js
@@ -1,4 +1,4 @@
-var parse = require('./parse');
+const parse = require('./parse');
 
 /**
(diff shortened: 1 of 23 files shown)


RobinCK:
@remy 
I think here [link](https://github.com/remy/nodemon/pull/1313/commits/c12208bf5a63154d7b25bf61607bf7575ad0da8d#diff-b2048f4188632273753b02bbcc2e31deL139) was a bug

remy:
@stablebot back off please. I'll review this PR.

remy:
IIRC node@6 doesn't need the `'use strict'` in the scripts? I'd like to include this PR in a general refactor release that also drops node@4 at the same time - which would allow me to drop the `'use strict'` in the files too. 

Just can't remember if that's right or not!

RobinCK:
Yes you are right

remy:
Gah. @stalebot! I'll re-open this and merge with the new linting rules in the coming weeks.

PR #1337, opened 2018-05-07
fix: ignore bare directories fully

### LINKED ISSUE #1336: nodemonConfig- ignore not working
version @1.17.4 - ignoring files in nodemonConfig not working in both package.json or nodemon.json

remy:
In the previous version, although the restart would ignore, an ignored
directory, if "bare" (i.e. without globs) would still be "monitored" and
potentially take up time. This change ensures the directory is fully
ignored (as well as a small change for fully ignoring files using the
same method).

Now the config.monitor rules is the source of truth for ignoring.

Fixes #1336

### DIFF
--- a/lib/monitor/match.js
+++ b/lib/monitor/match.js
@@ -55,6 +55,7 @@ function rulesToMonitor(watch, ignore, config) {
 
     try {
       var stat = fs.statSync(dir);
+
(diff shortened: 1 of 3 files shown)


remy:
Ping

Issue #1018, opened 2017-04-28
Spawn nodemon as child process?


nawgszy:
There is this casual little section in ```doc/events.md``` where nodemon is spawned as a child process:

```
var app = spawnNodemon();

app.on('message', function (event) {
  if (event.type === 'start') {
    console.log('nodemon started');
  } else if (event.type === 'crash') {
    console.log('script crashed for some reason');
  }
});

// force a restart
app.send('restart');

// force a quit
app.send('quit');
```

What is this "spawnNodemon" function? I am trying to perform async behaviour, so that before the "exec" statement listed above runs, I can also run my unit tests; however, some of them are asynchronous (as I am testing an API, for example). Accordingly, using the on('restart', handler) event, the 'exec' statement triggers before my async tests finish. See, I wish I could do the following with nodemon as a requireable:

```
nodemon({ exec: "npm run lint && npm start" })
.on('restart', (files, done) => {
  // returns an array of folders to recursively test with mocha
  foldersToTest = getFoldersToTest(files);

  // runs tests for these folders
  Promise.all(foldersToTest.map(runMochaTest))
  .then(done)
})
```

Obviously though, this is not feasible; but if I could do something like the following with the child process:

```
app.on('message', (event) => {
  if (event.type === 'start')
    nodemon({ exec: "npm run lint && npm start" }) // this seems wrong but there's an equivalent for app I am sure
  else if (event.type === 'restart')
    runTests().then(() => { app.send('start') })
})
```

I think I could make the async behaviour work. Of course, if I am understanding the result of spawnNodemon incorrectly, tell me, but otherwise I just need direction to what that function is. Thanks!

remy:
The spawnNode was supposed to be `fs.spawn`. but i can't remember, check the node docs for which spawn method allows for message events.

PR #1362, opened 2018-06-05
docs: more docs for nodemon as child process

andiwinata:
I was trying to find out how to use nodemon as child process, and stumble upon this https://github.com/remy/nodemon/issues/1018

So adding more docs for other people later to make it clearer how to do it.


remy:
👍 ❤️

PR #1380, opened 2018-07-06
Added "clear" command

KlaussMC:
Simply type "clear" into the terminal and the screen will be cleared.

### DIFF
--- a/README.md
+++ b/README.md
@@ -73,6 +73,8 @@ nodemon was originally written to restart hanging processes such as web servers,
 
 Whilst nodemon is running, if you need to manually restart your application, instead of stopping and restart nodemon, you can type `rs` with a carriage return, and nodemon will restart your process.
 
+To clear the screen, if ever you need to, type `clear` with a carriage return and nodemon will clear the screen.
+
 ## Config files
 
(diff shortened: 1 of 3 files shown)


remy:
Why not just hit ctrl+l?

KlaussMC:
Dear Remy Sharp

I Didn't know that was a thing, also I have never successfully contributed
to an open source project. This is a first for me and it was a feature that
I find quite handy.

Sincerely Jacob

KlaussMC:
I Didn't know that was a thing, also I have never successfully contributed
to an open source project. This is a first for me and it was a feature that
I find quite handy.

not to mention that it's cool AF for no other reason than being able to say *I made an edit to this awesome project that is really damn famous and useful.*

remy:
`ctrl+l` is a standard method to clear all terminal output (including things like your browser's console). It actually doesn't work in nodemon's output because nodemon isn't looking for that on `stdin`.

I think it makes sense to support `ctrl+l` in nodemon. However, the changes you've submitted don't actually do anything. Best case (that I can gather from your change) is that it will automatically clear the console when nodemon starts for the first time. It's not reading from stdin (which is suggested in the readme change), and it's currently failing tests - all reasons that I have to close this PR (sorry - as much as I love contributions, I do need to make sure changes are thorough).

Thanks so much for your contribution all the same.

KlaussMC:
It works for me, I am using it as we speak

just type clear into it, has to be lowercase

Issue #751, opened 2016-01-11
How to cleanly exit nodemon


edwardhorsford:
We have a situation where we would like to completely quit node and nodemon.

We are using `process.exit(0)` to cleanly exit node. When this happens, we also want nodemon to quit. Instead it stays running and monitoring the directory.

How do we cleanly exit nodemon from within the app?

NerdPad:
+1

Would like to know this as well.

dundalek:
I was looking for this too. I found a workaround that might be helpful.

If you exit with `process.exit(2)` it will cause nodemon also to exit. Nodemon will print stack trace so you might also hook into `error` event to make it clean, e.g.:
```js
nodemon(...)
  .on('error', function(code) {
    if (code !== 2) {
      console.log('Exited with error code: ', code);
    }
  });
```

if you don't want to or can't modify the child process it should be possible to run it like `your_command || exit 2` to make it work.

More details: https://github.com/remy/nodemon/issues/627

stevenvachon:
In tests, I cannot fully kill a nodemon process. `emit('quit')` only triggers the listener(s) and:

```js
nodemon(...).on('quit', function(){ this.reset() })
```
causes:
```
Assertion failed: (0), function uv_close, file ../deps/uv/src/unix/core.c, line 173.
```

joeytwiddle:
There has been some debate about removing the quit behaviour on exit code 2.

So I would not recommend this technique, if you can find an alternative way.

OoDeLally:
What is the state of this?
When `process.exit(2)` I get 
```
[nodemon] app crashed - waiting for file changes before starting...
```
then I have to `Ctrl+C` a second time to actually kill it.

1- Is it the expected behavior?
if yes, is a `app crashed` red message really appropriate? `Ctrl+C` is not a "crash" per se.
2- Any way to nicely quit everything in one key stroke (not twice) and without a scary crash message?
Thanks!

remy:
@OoDeLally exit(2) is a crash code (from your script), so if you're catching the exit code inside your program and using exit 2, then nodemon is behaving exactly as it should. 

A clean exit code is 0, anything above is a bad exit code, and nodemon will catch it. 

Also, this issue was raised 3 years ago, so is unrelated to your original question.

OoDeLally:
Thank you @remy. `process.exit(0)` indeed quit nicely the nodejs process, but not nodemon.
[...]

Issue #627, opened 2015-09-09
Is the special case for exit code 2 necessary? (JSHint integration)


pghalliday:
I see that exit code 2 has special meaning for bash (http://tldp.org/LDP/abs/html/exitcodes.html) however I am trying to use nodemon with JSHint. Unfortunately JSHint uses exit code 2 to signal that there are lint errors so nodemon exits the first time I forget a semi-colon with the following error.

```
[nodemon] failed to start process, possible issue with exec arguments
```

I can't actually see any reason nodemon should exit in this case, whether JSHint should be using that exit code or not

I have raised a corresponding issue at jshint/jshint#2672

pghalliday:
In case anyone lands here with the same problem my simple workaround for now is to call nodemon like this:

```
nodemon -e js -x 'jshint --exclude node_modules . || exit 1'
```

lukeapage:
the code is here:https://github.com/remy/nodemon/blob/e177dd7865376876b01c8c4b069bf21aaf75cf50/lib/monitor/run.js#L118

it was added with this commit:
https://github.com/remy/nodemon/commit/9651ab88b0e0a37a8d68dfbaf7ed6314993eb77d

lukeapage:
this is what node document as its exit codes: https://github.com/nodejs/node-v0.x-archive/blob/master/doc/api/process.markdown#exit-codes
and they purposefully avoid 2.

remy:
Okay, I think it makes more sense to change nodemon rather than hope for a breaking change to update in jshint. I can't remember exactly why I added this code, but I was being intentional. 

Do we think the entire condition should be removed based on the exit codes from node's docs?

lukeapage:
Well, it seems you do get warning 2 when its a bad command, but you can get error code 2 from any app for a variety of different reasons - I don't think you can assume its a bad command. Maybe log a warning? or add an option for fatalErrorCode?

Probably both nodemon should be more accepting or error code 2 and jshint shouldn't use it.

dallonf:
Seems that Mocha also throws error code 2. +1 for Nodemon accepting it as a normal failure.

deepwell:
I'd love to see this fixed as well. Mocha exits with code 2 when there are 2 errors (code 3 for 3 errors, etc.). That mocha "feature" defeats nodemon from continually watching and re-running the tests as I fix them.

dallonf:
Imo, this is still a problem. Unless this is an explicit "won't fix" from the devs, this should remain open.

[...]

PR #1381, opened 2018-07-10
Remove the special handling of exit code 2

### LINKED ISSUE #496: Correct way of using nodemon with test runners?
When running tests using mocha or the likes, if the tests fail, and thus the runner exits with a non-zero status code, nodemon seems to think there's something broken with the process and dies without watching for changes. Is there a way to opt out of this behavior?

joeytwiddle:
This PR removes the special handling of exit code 2.

Not only `mocha` but also `stylelint` will sometimes exit with code 2, resulting in unexpected behaviour from nodemon. (It exits, and stops monitoring.)

The ` || true` workaround is unusual and easy to forget. I only ever remember to add it *after* I have been bitten!

If there is [no special reason](https://github.com/remy/nodemon/issues/627#issuecomment-139053515) to keep the special handling of code 2, then I think it would be best to remove it.

(In the case when a developer does provide a non-parseable command, which exits with code 2, an error is already reported to the console. So we won't be keeping developers in the dark by not quitting. After this PR, the developer will need to hit Ctrl-C to exit nodemon, in order to fix their command. For an example syntax error, try: `nodemon -x 'ls <'`)

Fixes #496 #627

This could perhaps be part of a major release, just in case some people are using it as a way to break out of nodemon intentionally: https://github.com/remy/nodemon/issues/751#issuecomment-267431143

### DIFF
--- a/lib/monitor/run.js
+++ b/lib/monitor/run.js
@@ -158,6 +158,22 @@ function run(options) {
       process.exit();
     }
 
+    if (code === 2 && Date.now() < config.lastStarted + 500) {
+      // something wrong with parsed command
+      utils.log.error('process failed, unhandled exit code (2)');
+      utils.log.error('');
+      utils.log.error('Either the command has a syntax error,');
+      utils.log.error('or it is exiting with reserved code 2.');
+      utils.log.error('');
+      utils.log.error('To keep nodemon running even after a code 2,');
+      utils.log.error('add this to the end of your command: || true');
+      utils.log.error('');
+      utils.log.error('nodemon will stop now so that you can fix the command.');
+      utils.log.error('');
+      bus.emit('error', code);
+      process.exit();
+    }
+
     // In case we killed the app ourselves, set the signal thusly
     if (killedAfterChange) {
       killedAfterChange = false;
--- a/lib/monitor/run.js
+++ b/lib/monitor/run.js
@@ -158,8 +158,10 @@ function run(options) {
       process.exit();
     }
 
+    // If the command failed with code 2, it may or may not be a syntax error
+    // See: http://git.io/fNOAR
+    // We will only assume a parse error, if the child failed quickly
     if (code === 2 && Date.now() < config.lastStarted + 500) {
-      // something wrong with parsed command
       utils.log.error('process failed, unhandled exit code (2)');
       utils.log.error('');
       utils.log.error('Either the command has a syntax error,');
@@ -167,6 +169,7 @@ function run(options) {
       utils.log.error('');
       utils.log.error('To keep nodemon running even after a code 2,');
       utils.log.error('add this to the end of your command: || exit 1');
+      utils.log.error('');
       utils.log.error('Read more here: https://git.io/fNOAG');
       utils.log.error('');
       utils.log.error('nodemon will stop now so that you can fix the command.');
--- a/lib/monitor/run.js
+++ b/lib/monitor/run.js
@@ -158,13 +158,6 @@ function run(options) {
       process.exit();
     }
 
-    if (code === 2) {
-      // something wrong with parsed command
-      utils.log.error('process failed, unhandled exit code (2)');
-      bus.emit('error', code);
-      process.exit();
-    }
-
     // In case we killed the app ourselves, set the signal thusly
     if (killedAfterChange) {
       killedAfterChange = false;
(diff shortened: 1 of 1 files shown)


remy:
I'm not convinced this is the right change. I think the messaging in the exit should be clearer, but that's all.

With your changes, here is the workflow:

```
❯ nodemon -x 'ls <'
[nodemon] to restart at any time, enter `rs`
[nodemon] watching: *.*
[nodemon] starting `ls <`
sh: -c: line 0: syntax error near unexpected token `newline'
sh: -c: line 0: `ls <'
[nodemon] app crashed - waiting for file changes before starting...
```

Since the actual exec command is wrong, the developer has to `ctrl+c` nodemon and correct the exec command.

Automatically restarting on file changes will never make that command work.

I think the right change is to update the error message that's shown, perhaps linking to the FAQ (which can be added to with links and even more context).

I can't see a reason for nodemon _not_ to exit on code 2 - unless I'm missing something?

joeytwiddle:
The argument not to exit on code 2 is that mocha will sometimes exit with code 1, 2, 3, 4, ... (and stylelint too), and usually in those cases we still want `nodemon` to keep watching and running.

Currently behaviour with mocha is somewhat inconsistent.  `nodemon` will happily watch and run passing and failing tests for a while, but then it will surprisingly stop running entirely (when exactly 2 mocha tests fail). It appears random, until you find the GitHub issue explaining that code 2 is special, and you are advised to put `|| true` at the end of your mocha command.

On the other hand, yes sometimes code 2 means you wrote a bad shell command. And in that case exiting nodemon is slightly preferable.

So I guess we have to decide which is more bothersome: nodemon stops running mocha tests unexpectedly, or having to hit <kbd>Ctrl-C</kbd> after typing a bad command.

Anyway I do agree, in either case, some messaging may help to guide the developer. I will look into that.

Other options:
1. We could try harder to detect whether the code 2 is from a bad command or from mocha/stylelint. For example:
1a. we could read stderr for things like `sh: -c: line 0: syntax error`, or
1b. we could measure if the command fails _immediately_ the _first time_ it is executed.
2. We could raise issues on mocha and stylelint packages, explaining that they are using the reserved code 2 incorrectly. (But I suspect they will not want to change the behaviour.)

I would like to look into 1b, and situational messaging, when I have time.

joeytwiddle:
OK, I have changed the behaviour:
- If the exit code is 2, and the time taken for the child to fail is < 500ms, then we assume the command had a syntax error.
- I have improved the messaging a bit (perhaps verbose, but hopefully clear).
- (I have not checked if this is the first run or not. It doesn't seem necessary.)

To test, try:
```sh
./bin/nodemon.js -x 'ls <'

# and

./bin/nodemon.js -x 'sleep 1 ; exit 2'
```

lib/monitor/run.js
```diff
@@ -158,9 +158,18 @@ function run(options) {
...
-    if (code === 2) {
+    if (code === 2 && Date.now() < config.lastStarted + 500) {
       // something wrong with parsed command
       utils.log.error('process failed, unhandled exit code (2)');
+      utils.log.error('');
+      utils.log.error('Either the command has a syntax error,');
+      utils.log.error('or it is exiting with reserved code 2.');
+      utils.log.error('');
+      utils.log.error('To keep nodemon running even after a code 2,');
+      utils.log.error('add this to the end of your command: || true');
```
remy: Can you add: `Read more here: https://git.io/fNOAG` after line 169.

remy:
Two comments I've left, and once these are in, I'm happy with the PR - thank you and great work 👍

joeytwiddle:
Remy, I only saw one comment from you! I have added the link to the FAQ.

But what was the second comment? I have taken a guess, and changed the `|| true` into `|| exit 1` like the FAQ suggests.

Thank you for `nodemon`, an invaluable productivity tool! :smile:

remy:
I don't understand what github is doing, but I've taken a screenshot of the comments - sorry, bit naff!

screen shot 2018-07-17 at 17 00 04

joeytwiddle:
My apologies. I probably made a rebase/amend and a force push yesterday, just after you commented, so the comment got lost.

I have added a comment with the fNOAR link now.

remy:
> If you want to make adjustments to this branch directly, I will not be offended. 🙇

I've never really known how do that cleanly. Once I clone the branch locally, (from past experience) I couldn't push back into the PR and had to make my _own_ branch and new PR - really odd. Anyway, this looks good - thanks!

PR #1392, opened 2018-07-13
Fixed Unusual Backspace behaviour

KlaussMC:
when pressing backspace, the cursor would move backwards, but not delete the characters. It would also move beyond the beginning of the line and jump to the line above it. That is now fixed.

### DIFF
--- a/README.md
+++ b/README.md
@@ -2,11 +2,9 @@
 
 # nodemon
 
-For use during development of a node.js based application.
+nodemon is a tool that helps develop node.js based applications by automatically restarting the node application when file changes in the directory are detected. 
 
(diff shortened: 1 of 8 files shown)


remy:
It's highly likely that I'm going to completely revert the change that added support for ctrl-l due to the issues it's caused (I can't quite understand them yet), but for me to view this pull request, can you rebase it against the current master (otherwise it's showing a tonne of changes atm)

KlaussMC:
Sorry, rebase against the current master? do you mean to a different branch?

remy:
Sorry, I had to completely revert the way stdin was being handled, so this PR isn't required anymore. Thanks again.

PR #1401, opened 2018-08-11
docs: update help.txt

mef:
Small update so that `nodemon -h` mentions that `.json` files are now watched by default.

P.S. The change fixing #643 may have been considered as a breaking change, and reflected by bumping major version number. Updating some apps to above nodemon v1.7.0 require adaptation of the config...
P.P.S. Thanks for creating nodemon, super useful piece of software.


remy:
Thank you 👍

PR #1413, opened 2018-08-29
chore(cliDoc): Adding delay on help command output

willmendesneto:
Adding delay on help command output


remy:
That's already in the cli docs: https://github.com/remy/nodemon/blob/master/doc/cli/options.txt

PR #1441, opened 2018-10-26
Fix nodemon installation for node 10.x and beyond

eilgin:
Today i was preparing our projects for the LTS release of Node.js 10 but was stopped as your package is using a version of chokidar that was using a dependency (upath) which prevents any version beyond 9 (https://github.com/anodynos/upath/blob/v1.0.2/package.json#L43) so i've just bumped chokidar to the first version that has fixed this (https://github.com/paulmillr/chokidar/commit/11aea72ad7c65216a0f767142ce57151c7b3d0c1).

This issue will happen to anybody that is using a lockfile (specifically `yarn.lock`) that references any version of `nodemon` up to the current release and switch to node 10 (which could happen more and more often after the LTS release 😉).

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -302,9 +302,9 @@
       "integrity": "sha1-GdOGodntxufByF04iu28xW0zYC0="
     },
     "atob": {
-      "version": "2.1.1",
-      "resolved": "https://registry.npmjs.org/atob/-/atob-2.1.1.tgz",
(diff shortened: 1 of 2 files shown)


eilgin:
it's weird that travis has failed has i launch mine locally using `TRAVIS=1 npm test` and `echo $?` returns **0** 😕.

eilgin:
Wow, that was quick!

 👍

PR #1454, opened 2018-11-21
Remove pstree.remy

FallingSnow:
This is a bloated dependency that (down the stream) depends on a dependency that is [running unknown and clearly malicious code on everyone's system](https://github.com/dominictarr/event-stream/issues/116).

### DIFF
--- a/lib/monitor/run.js
+++ b/lib/monitor/run.js
@@ -12,7 +12,7 @@ var child = null; // the actual child process we spawn
 var killedAfterChange = false;
 var noop = function () { };
 var restart = null;
-var psTree = require('pstree.remy');
+var killTree = require('tree-kill');
 var path = require('path');
 var signals = require('./signals');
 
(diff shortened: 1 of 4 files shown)


jaydenseric:
Please lets expedite a review and merge! It's unbelievable a hack affecting millions of installations is not being acknowledged or dealt with by maintainers 🤯

remy:
_maintainer_ - singular. 

You've commented out a bunch of tests. Can you explain why? (I don't know why the PR is on nodemon and not my pstree, but I'll review the change to make sure the swapped in module does the same)

remy:
You've also bumped the semantic release several major versions, and the auto deployment depends entirely on that code - and has not been tested with major versions out...

Bulking together these changes is delaying the release.

remy:
Right - reviewing this properly, which I'm guessing never happened, this can't be merged.

The reason I'm using pstree.remy - my own fork, is (as it clearly states in the description of the module) is that it's able to get the process tree *without* `ps` - which a bunch of systems (usually inside docker) use. `kill-tree` relies on `ps` so this would break again.

The fix is to PR against pstree.remy and to remove the `ps-tree` dep _there_.

remy:
`npm audit` is great, except it's kinda not. It covers dev deps which are a very different kind of dependency than production (which is why I use snyk instead).

I'm pushing a release now that removes the deep dep of ps-tree (fixed in pstree.remy).

PR #1472, opened 2018-11-28
feat: Send SIGHUP to application for graceful restart

gja:
This feature can be used for applications which are already
configured to catch a SIGHUP, and reload their configuration.
Unlike --signal, this does not propagate the signal to all
processes in the process tree, or unpipe stdin.

The command line option --sighup supports this feature

### DIFF
--- a/README.md
+++ b/README.md
@@ -2,7 +2,7 @@
 
 # nodemon
 
-nodemon is a tool that helps develop node.js based applications by automatically restarting the node application when file changes in the directory are detected. 
+nodemon is a tool that helps develop node.js based applications by automatically restarting the node application when file changes in the directory are detected.
 
 nodemon does **not** require *any* additional changes to your code or method of development. nodemon is a replacement wrapper for `node`, to use `nodemon` replace the word `node` on the command line when executing your script.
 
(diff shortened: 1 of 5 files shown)


remy:
Can you explain why using `--signal SIGHUP` wouldn't work instead of adding a new feature to nodemon?

I'm guessing it's for when you use node in a cluster, and the signal is reaching all the worker processes of the master node process? If that's the case, shouldn't the workers be catching and ignoring the SIGHUP, and the master implementing your `reloadConfig`?

gja:
There are two reasons
1) I didn't actually notice the earlier --signal until much later, so I really wanted to update the documentation. The docs do mention how to capture the event, but it's unclear if nodemon is waiting for the process to die.
2) This isn't how systems such as `upstart` behave. AFAIK, in a typical daemon, the SIGHUP signal is to be sent to the parent process, and the parent is responsible for figuring out how to communicate the change to the children. Often this may be by broadcasting the signal, or just by setting a flag marking processes to be killed after their current job (apache does the latter).

Especially when you consider that the process tree may not end with your application (and somewhere along the process tree you might have spawned a background job)

remy:
Nodemon does try to wait for the master process to finish, but it also has a timeout (and there's also times, quite often, when nodemon just can't tell if the process has finished).

As for not sending kill signals to the worker processes - as much as technically I agree, the main use case is to have nodemon shutdown *all* the workers. This is most apparent when you'll have (what's a fairly typical) setup of having nodemon start an npm run script which fires up ts-node or babel-node which fires up another sub process.

When nodemon shutdowns the main process, it needs to make sure all the worker processes also shutdown.

---

In general (as per the readme) I try to avoid adding more flags and options to nodemon, and I believe in this case there's a solid practice to solving the problem using nodemons existing infrastructure. 

I don't think an update to the faq would go amiss though if you'd like to include anything there?

remy:
I'm going to close this PR. I fully appreciate PRs and they're a big deal - so thank you. There's a decent solution that can use nodemon's existing arch, and I'll update the FAQ with what I've posted here.

Thanks again.

Issue #1528, opened 2019-02-14
Nodemon does not reload on --watch folder only on the app file


MrDesjardins:
- `nodemon -v`: 1.18.10
- `node -v`: 10.8.0
- Operating system/terminal environment: MacOS, bash command
- Using Docker? What image: No
- Command you ran: `nodemon --watch ./build/dist --watch ./build/dist/graphql/schemas build/dist/index.js -V`

### Expected behavior
To have NodeJS restart.

### Actual behavior
It detects the folder has a file that has changed but does not restart.

### Steps to reproduce
1) NPM script:
``` 
"watchgraphql": "cpx 'src/graphql/schemas/**/*.graphql' build/dist/graphql/schemas/ -w -v",
"dev": "concurrently \"tsc -w\" \"npm run watchgraphql\" \"nodemon --watch ./build/dist --watch ./build/dist/graphql/schemas build/dist/index.js -V\"",
``` 
2) Run the script
```
npm run dev
``` 

This check for TypeScript file, when something is saved, transpile and save the JS. This change the index.js thus node is restarting as expected. However, if I change a .graphql file, the `watchgraphql` copy the file (as expected) and `nodemon` detect the change (see the verbose output) but node does not restart.

Verbose output:

> [1] Copied: src/graphql/schemas/ocadmin/org.graphql --> build/dist/graphql/schemas/ocadmin/org.graphql
> [2] [nodemon] files triggering change check: build/dist/graphql/schemas/ocadmin/org.graphql
> [2] [nodemon] matched rule: /Users/pdesjardins/code/ocgraphqlserver/build/dist/graphql/schemas/**/*
> [2] [nodemon] changes after filters (before/after): 1/0
> [2] [nodemon] files triggering change check: build/dist/graphql/schemas/ocadmin/org.graphql
> [2] [nodemon] matched rule: /Users/pdesjardins/code/ocgraphqlserver/build/dist/graphql/schemas/**/*
> [2] [nodemon] changes after filters (before/after): 1/0
> [2] [nodemon] files triggering change check: build/dist/graphql/schemas/ocadmin/org.graphql
> [2] [nodemon] matched rule: /Users/pdesjardins/code/ocgraphqlserver/build/dist/graphql/schemas/**/*
> [2] [nodemon] changes after filters (before/after): 1/0

zhylmzr:
same problem.
I found that the change of typescript file doesn't cause a restart.

zhylmzr:
> same problem.
[...]

PR #1537, opened 2019-03-06
fix: to avoid confusion like in #1528, always report used extension

bleadof:
PR based on https://github.com/remy/nodemon/issues/1528#issuecomment-470149782


remy:
Took me a while to realise this was causing so many people some brain ache. Thanks for the PR, nice and simple. Merging and releasing now.

PR #1551, opened 2019-04-10
Quote zero-length strings in arguments

forivall:
If a zero-length string is passed, it does not get properly quoted, and then it is not properly passed to the child process

I noticed this issue when using nodemon to monitor a wrapper script, where i basically pass a bunch of arguments like function arguments.

TODO: update the commit message to fit your convention. I just did this quickly in the github editor.

### DIFF
--- a/lib/utils/index.js
+++ b/lib/utils/index.js
@@ -69,7 +69,7 @@ var utils = (module.exports = {
       args.map(function (arg) {
         // if an argument contains a space, we want to show it with quotes
         // around it to indicate that it is a single argument
-        if (arg.indexOf(' ') === -1) {
+        if (arg.length > 0 && arg.indexOf(' ') === -1) {
           return arg;
         }
         // this should correctly escape nested quotes


remy:
I'll fix this commit message on merge.

What was the case that caused this to be a bug (out of interest)?

forivall:
I have a tool that shells out to launch a wrapper around an api, where, to be lazy, I just use the position of arguments rather than any kind of arg parsing. I pass empty args to indicate that they're not set. The code looks like this:
```ts
  const args = [pkg, argv.entryFile || '', argv.api || '', argv.configFile || '']
  const nodemonConfig = {
    script: resolve('.runner-build/index.js'),
    args,
    stdout: true,
    delay: 0.5,
    ignore,
  }

  if (argv.runnerWatch) {
    nodemonConfig.watch = ['.', resolve('.runner-dist')]
  }

  const n = nodemon(nodemonConfig)
```
Hope that makes sense!

PR #1552, opened 2019-04-10
feat: add TypeScript to default execPath

leonardodino:
`ts-node` is the standard for running typescript node programs on development mode.

Adding this line will enable everyone with a `tsconfig.json` to have a full-refresh server watching experience. (:

### DIFF
--- a/lib/config/defaults.js
+++ b/lib/config/defaults.js
@@ -7,6 +7,7 @@ module.exports = {
   execMap: {
     py: 'python',
     rb: 'ruby',
+    ts: 'ts-node',
     // more can be added here such as ls: lsc - but please ensure it's cross
     // compatible with linux, mac and windows, or make the default.js
     // dynamically append the `.cmd` for node based utilities


remy:
Not stale.

remy:
@teebot hmm, this seems rather important. Can you raise a separate issue (allowing others to find the problem).

It seems like we'll either need to be smart about adding ts support or removing it…

mefcorvi:
Well, this pull request explains why now I'm getting `Unknown or unexpected option: --inspect` error. The `ts-node` doesn't support that option and strictly speaking `ts-node` is not a drop-in replacement for `node`, in some cases you need to run `node` for ts files.

leonardodino:
sorry guys, my bad.

> You still use node -r ts-node/register app.ts in production.
I didn't knew people were running `ts-node` in production, I always compile first and use it only for development.

leonardodino:
as of the last release, you can now replace:
`nodemon -r ts-node/register app.ts` => `nodemon app.ts`

mefcorvi:
> and I think the equivalent for custom node options will be overriding the ts-node

I've solved my issue by overriding executable for "ts" files back to `node` in nodemon configurations :-) After all explicit is better than implicit. Not sure if I must close #1565 I think that it's worth to mention somewhere in docs that `ts` files are processed by `ts-node` after 1.19.0.

leonardodino:
Good one @mefcorvi (: Sorry for causing trouble.

I think we agree that having `ts-node` can help on the default scenarios, and overriding it when required by project needs.

I agree that it should be documented somewhere, especially as it's a recent change in a long-running project.

backbone87:
can this either be properly documented or reverted? it cost me several hours to figure out why changing from: `node -r ts-node/register ./src/index.ts` to `nodemon ... -r ts-node/register ./src/indexts` didnt work. the ultimate reason is, that ts-node will be executed twice now and the 2nd one (from the cmd flag `-r ts-node/register`) will see the transpiled files already and complain that there are implicit types (because types are already compiled away by first ts-node instance from the implicit `--exec ts-node`).

another solution would be to check the arguments, when using implicit `--exec ts-node`, and remove any `-r|--require ts-node/register[/transpile-only]`

HosseinAgha:
@remy this broke lots of existing codebases that use `-r ts-node/register` in their `dev` script!

remy:
@HosseinAgha I raised another issues god knows when, saying that as I don't use typescript I really don't know what the right way of handling or fixing this. I've wondered for a long time if this change should be reverted (indeed asking so), but again, I don't use typescript so I've no idea.

Someone, please take a lead on what's right here and I'll happily merge in if it gets the consensus.

backbone87:
@remy the correct approach would be to only set the impicit `--exec ts-node`, if a) the script argument is a `.ts` file **and** b) there is no `-r|--require ts-node/register[/transpile-only]` in the options passed through to the executeable.

mrozekma:
Thought I'd mention that this also breaks `esm`, as it has to be loaded before `ts-node`; the normal pattern when combining it with Typescript is:

    node -r esm -r ts-node/register main.ts

Using `nodemon` instead of `node` here obviously doesn't work, and even if you take out `-r ts-node/register` it breaks because the ordering is wrong

Issue #1476, opened 2018-12-01
Nodemon Fails to Wait for Process Exit when using Docker + Inspector


darkobits:
> Created per @remy's request in [this issue](https://github.com/remy/nodemon/issues/1346#issuecomment-442367096).

It seems that when restarting a process, Nodemon isn't waiting for the old process to completely exit before starting a new one, but:

- _only_ when in a Docker container and
- _only_ when the inspector/debug mode is active.

This issue can be intermittent in applications without a shutdown handler. But, having an application that implements a listener for Nodemon's `SIGUSR2` signal that takes even a second or so to complete can be enough to cause this issue to happen reliably.

The presence of `--inspect` _does_ cause the new process to immediately crash because its trying to open a debug server on the same port that is still in use by the previous process' debug server, but this merely a side-effect of this issue.

If running Nodemon in a Docker container without `--inspect`, Nodemon correctly waits for the old process to exit, but when `--inspect` is used inside the container, Nodemon _does not_ wait for the process to exit, leading to an immediate crash.

This can be reproduced with Nodemon `1.18.7` and the following:

```
$ node -v
v10.13.0

$ docker -v
Docker version 18.09.0, build 4d60db4

$ docker-compose -v
docker-compose version 1.23.1, build b02f1306
```

Here is a repo that replicates this issue: https://github.com/darkobits/nodemon-restart-issue.

ImTheDeveloper:
Just started to see this issue when I upgraded to latest versions of nodemon too.

remy:
I think this _might_ be solved with the next release going up [at the moment](https://travis-ci.org/remy/nodemon/builds/466125111) - assuming it passes, it'll land in nodemon@1.18.8.

@darkobits can you test this once it's up?

denic:
@remy Works for me. Thanks!

ImTheDeveloper:
Great thank you for this fix!

bighuggies:
I'm still seeing the issue with `1.18.8`

daniele-orlando:
`v1.18.8`

Same here, less frequent but still there.

darkobits:
I've updated the [demo repo](darkobits/nodemon-restart-issue) referenced in the original post to nodemon `1.18.9`, and this issue is still occurring.

darkobits:
@remy, thoughts regarding leaving this closed vs. re-opening?

justinpage:
[...]

PR #1579, opened 2019-06-07
fix: wait for all child processes to terminate (fixes issue #1476)

pepe79:
See #1476

### DIFF
--- a/lib/monitor/run.js
+++ b/lib/monitor/run.js
@@ -16,6 +16,28 @@ var psTree = require('pstree.remy');
 var path = require('path');
 var signals = require('./signals');
 
+function waitForSubprocesses(subprocesses, callback) {
+  if (Array.isArray(subprocesses) && subprocesses.length > 0) {
+    // check if all old subprocesses have been terminated
+    exec('kill -0 ' + subprocesses.join(' '), (error) => {
+      const returnCode = error ? error.code : 0;
+      if (returnCode < 126) { // ignore command not found error
+        const stillRunning = subprocesses.length - returnCode;
+        if (stillRunning > 0) {
+          utils.log.status('still waiting for ' + stillRunning +
+            ' subprocess(es) to finish...');
+          setTimeout(waitForSubprocesses.bind(this, subprocesses, callback),
+            100);
+          return;
+        }
+      }
+      callback();
+    });
+  } else {
+    callback();
+  }
+}
+
 function run(options) {
   var cmd = config.command.raw;
 
@@ -341,17 +363,24 @@ function kill(child, signal, callback) {
     // configured signal (default: SIGUSR2) signal, which fixes #335
     // note that psTree also works if `ps` is missing by looking in /proc
     const sig = signal.replace('SIG', '');
-    psTree(child.pid, function (err, kids) {
+    psTree(child.pid, function (err, subprocesses) {
       if (psTree.hasPS) {
-        spawn('kill', ['-s', sig, child.pid].concat(kids))
-          .on('close', callback);
+        spawn('kill', ['-s', sig].concat(subprocesses))
+          .on('close', waitForSubprocesses.bind(this, subprocesses,
+            function () {
+              spawn('kill', ['-s', sig, child.pid])
+                .on('close', callback);
+            }));
       } else {
         // make sure we kill from smallest to largest
-        const pids = kids.concat(child.pid).sort();
+        const pids = subprocesses.slice().sort();
         pids.forEach(pid => {
           exec('kill -' + signals[signal] + ' ' + pid, () => { });
         });
-        callback();
+        waitForSubprocesses(subprocesses, function () {
+          exec('kill -' + signals[signal] + ' ' + child.pid, () => { });
+          callback();
+        })
       }
     });
 


remy:
Is there a test for this that shows it failing and then passing with the change?

remy:
Also, how is this different to the current kill logic?

pepe79:
> Also, how is this different to the current kill logic?

The current kill logic is unchanged. After issuing the kill command in run.js:376 (which only sends signals to the other processes), the processes may need some time finishing.
The new command `kill -0` actually does not send a signal, it only reports through the exit code, wether some given PIDs still exists, i.e. the process has not terminated. The new logic waits until every child process is really terminated. The current logic only waits for the parent child process (the sh command) to finish (the 'exit' event).

remy:
And I assume this doesn't apply to windows? (I see Windows is being skipped)

pepe79:
> And I assume this doesn't apply to windows? (I see Windows is being skipped)

As mentioned in my comment in #1476 I assume that the issue is only occuring on linux platform.

axmor-b-common:
@pepe79 Could you please look through the Travis error? I'm not sure whether it's possible to fix it otherwise.

Cerberuser:
@pepe79 Could you please look through the Travis error? I'm not sure if it is possible to fix it otherwise.

pepe79:
@Cerberuser yes I will have a look within the next 2 days.

dobesv:
I think this may address the issue but it spams the console with messages "waiting for 1 child process to exit".

In our system we actually never exit when we receive the signal, we do a hot reload and ignore the signal.  I think nodemon should allow this without spamming the console or using a lot of CPU cycles.

pepe79:
hey @dobesv , ok that sounds interesting. I did not thought about this use-case. It's conceptually the opposite of what is addressed in the issue #1476. I don't think that both use-cases
1) wait until the process has really terminated and then restart it
and
2) custom signalhandler without terminating the process 
work together without a specific command line switch that tells nodemon what to do.

@remy  What do you think about theses use-cases?

lib/monitor/run.js
```diff
@@ -343,15 +364,21 @@ function kill(child, signal, callback) {
...
-          .on('close', callback);
+        spawn('kill', ['-s', sig].concat(kids))
+          .on('close', waitForOldKids.bind(this, kids, function () {
+            spawn('kill', ['-s', sig, child.pid])
+              .on('close', callback);
+          }));
       } else {
         // make sure we kill from smallest to largest
-        const pids = kids.concat(child.pid).sort();
+        const pids = kids.sort();
```
dobesv: You are mutating your parameter here, which is kind of a no-no.  Perhaps copy the array first.

lib/monitor/run.js
```diff
@@ -16,6 +16,27 @@ var psTree = require('pstree.remy');
 var path = require('path');
 var signals = require('./signals');
 
+function waitForOldKids(oldKids, callback) {
+  if (oldKids) {
```
dobesv: Maybe also check `oldKids.length > 0` ?

lib/monitor/run.js
pepe79: You're right. I fixed that by using `.slice()` now.

lib/monitor/run.js
```diff
@@ -343,15 +364,21 @@ function kill(child, signal, callback) {
     const sig = signal.replace('SIG', '');
     psTree(child.pid, function (err, kids) {
       if (psTree.hasPS) {
-        spawn('kill', ['-s', sig, child.pid].concat(kids))
-          .on('close', callback);
+        spawn('kill', ['-s', sig].concat(kids))
+          .on('close', waitForOldKids.bind(this, kids, function () {
```
dobesv: If we filter out `child.pid` from `kids` here, we can just rely on the `exit` event from `ChildProcess` to detect the root process exit, which avoids the extra polling in the case that there's no extra child processes.  e.g. `kids.filter(p => p !== child.pid)`.

dobesv:
The current release of nodemon supports both scenarios perfectly, as long as there is only one child process.  I think a few simple changes will make your change more compatible with this:

1. Poll the children less often to avoid heavy CPU use (once per second is probably fine)
2. Don't poll the `child.pid`, since node will notify us of its exit via the event system.  If there are no extra child processes the polling won't run at all, and your change won't affect users who don't need it.
3. Don't log on every poll; maybe just log if the number of remaining processes changed from the last time (or don't log at all)

lib/monitor/run.js
```diff
@@ -16,6 +16,27 @@ var psTree = require('pstree.remy');
 var path = require('path');
 var signals = require('./signals');
 
+function waitForOldKids(oldKids, callback) {
+  if (oldKids) {
+    // check if all previous kids have been terminated
+    exec('kill -0  ' + oldKids.join(' '), (error) => {
+      const returnCode = error ? error.code : 0;
+      if (returnCode < 126) { // ignore command not found error
+        const stillRunningKids = oldKids.length - returnCode;
```
dobesv: When I run `kill` it always returns `0` or `1` so I'm not sure if this math is reliable.  You might have to run psTree again for each child to see whether it is still running and whether it has added any new child processes.

lib/monitor/run.js
```diff
@@ -16,6 +16,27 @@ var psTree = require('pstree.remy');
...
+        if (stillRunningKids > 0) {
+          utils.log.status('still waiting for ' + stillRunningKids +
+            ' child process(es) to finish...');
+          setTimeout(waitForOldKids.bind(this, oldKids, callback), 100);
```
dobesv: A slower polling interval might be nice for processes that aren't actually planning to exit.

lib/monitor/run.js
dobesv: Maybe what you could do is just poll `kids[0]`.  If it isn't running any more, call `waitForOldKids(oldKids.slice(1), callback)`.  Then when `oldKids.length === 0` you are done.

dobesv: Also, using `psTree` to wait for the child process might work in windows as well, where `kill` is not available.

lib/monitor/run.js
```diff
@@ -343,15 +364,21 @@ function kill(child, signal, callback) {
...
+          }));
       } else {
         // make sure we kill from smallest to largest
-        const pids = kids.concat(child.pid).sort();
+        const pids = kids.slice().sort();
         pids.forEach(pid => {
           exec('kill -' + signals[signal] + ' ' + pid, () => { });
         });
-        callback();
+        waitForOldKids(kids, function () {
```
dobesv: Here, too, `kids.filter(pid => pid !== child.pid)` might be nice to avoid polling `child.pid` since we'll get an event for that.

lib/monitor/run.js
pepe79: Ok, changed this line to `if (Array.isArray(oldKids) && oldKids.length > 0) {`

lib/monitor/run.js
```diff
@@ -343,15 +364,21 @@ function kill(child, signal, callback) {
...
-        spawn('kill', ['-s', sig, child.pid].concat(kids))
-          .on('close', callback);
+        spawn('kill', ['-s', sig].concat(kids))
+          .on('close', waitForOldKids.bind(this, kids, function () {
+            spawn('kill', ['-s', sig, child.pid])
+              .on('close', callback);
+          }));
       } else {
         // make sure we kill from smallest to largest
-        const pids = kids.concat(child.pid).sort();
```
dobesv: I'm assuming you removed the `concat` here because `psTree` already includes the given `child.pid` in the list.

lib/monitor/run.js
pepe79: I think I know what you're meaning. But some commands won't create any sub processes. Maybe it is esoteric but for example `nodemon --exec  "read VAR"` won't start a subprocess because `read` is a shell builtin command.

lib/monitor/run.js
```diff
@@ -343,15 +364,21 @@ function kill(child, signal, callback) {
     const sig = signal.replace('SIG', '');
     psTree(child.pid, function (err, kids) {
       if (psTree.hasPS) {
-        spawn('kill', ['-s', sig, child.pid].concat(kids))
-          .on('close', callback);
+        spawn('kill', ['-s', sig].concat(kids))
+          .on('close', waitForOldKids.bind(this, kids, function () {
+            spawn('kill', ['-s', sig, child.pid])
```
dobesv: It seems like we already sent the signal to `child.pid` above, and then do it again here.

lib/monitor/run.js
dobesv: `read VAR` will start a shell subprocess, although there will be just one subprocess from nodemon's perspective.  But in that case the `exit` event will be sent to us for that process so we don't need to poll for it to exit.

pepe79:
> The current release of nodemon supports both scenarios perfectly, as long as there is only one child process. 

I think your assumption is wrong, because nodemon itself starts a `sh` process, which then starts the provided `exec` command. So the "normal" usage of nodemon with launching one simple command, launches at least two processes. The exit-handler of nodemon only handles the exit of the `sh` command which does not gurantee that the subprocess (the exec command) also has terminated (at least on linux).

lib/monitor/run.js
dobesv: Or to clarify what I'm trying to say.  `nodemon --exec "read VAR"` will at least start `/bin/sh` to run the `read` built-in.  So there's still a child process there, but only one (`/bin/sh`).  `nodemon --exec "python foo.py"` runs two subprocesses - `/bin/sh` and `python`.

dobesv:
> > The current release of nodemon supports both scenarios perfectly, as long as there is only one child process.
> 
> I think your assumption is wrong, because nodemon itself starts a `sh` process

It only does this if you aren't using node without any special arguments.

If you run `nodemon bla.js` it doesn't run `sh`, it just forks a new `node` process and passes the script to it directly.  In that case the bug(s) you are trying to fix here do not manifest because there is only a single child process.

Currently these bugs don't affect me because I only ran into them while trying pass arguments to `node` when running with `nodemon`.  This triggered it to use `sh` as an intermediate process to pass the arguments.  To resolve this I now pass arguments to `node` via environment variables instead of arguments through `nodemon`.  Thus this issue does not currently affect me.

Your fix, were it to be merged, might affect me if it changes the behavior of the case that currently works correctly for me (single child process).

pepe79:
Ah ok I see. Thank you for clarification. My command gets spawned and not forked.

lib/monitor/run.js
pepe79: Ok, you think `kids` contains also the `sh` process PID. I thought it doesn't (because of the previous old code in line 346: `..., child.pid].concat(kids)`). I have to clarify what is right here.

dobesv: Hmm actually I might have been confused about this one, I did a quick test and it seems like `pstree` is not including the parent id in the result.

e.g.

```
> child = require('child_process').exec('sleep 300s')
ChildProcess {
> child.pid
22955
> require('pstree.remy')(child.pid, (e, k) => console.log(k))
undefined
> [ 22956 ]

```

dobesv: I deleted some of my erroneous comments about this...

lib/monitor/run.js
dobesv: Actually since `psTree` doesn't return the pid given as part of the list, you can't use it to tell whether a process is still running.

pepe79: In this part of code I only want to wait for the subprocesses to terminate. I think using pstree here instead of `kill -0` is a good idea.

dobesv: Oh yeah, if you pass in the root pid you can just us pstree to check if there are any children or not instead of trying to kill each child.

Cerberuser:
Is this now mainly the problem of @dobesv's use-case, or we should just wait for @remy? The problem in question arises fairly regularly, so we highly appreciate the solution.

remy:
Question: is it possible to add a test that shows this solution works? (not intended as a snarky question! genuinely - tests can be tricky with nodemon as the rely on spawning processes within tests and shutting them down).

uptownhr:
Any progress on developing a test to show this works? Currently breaking nodemon for us.

hschmiedhofer:
> Any progress on developing a test to show this works? Currently breaking nodemon for us.

Are there any alternatives for nodemon?

dobesv:
You could probably work around this issue by making a is file that runs the command you want to run, but waits for it to exit before exiting itself.  Basically replace the sh process with your own node process that handles the signal from nodemon.

If the process you are running is already a node process, use environment variables to pass node options instead of passing them to nodemon.

dobesv:
You could try "forever" as an alternative.  I'm sure there are others out there.

Cerberuser:
@pepe79, is there any progress, or the PR is really stale?

pepe79:
No progress from my side, because of lack of time. Next steps will be looking at some of the findings of @dobesv and thinking about how to create a testcase that shows the errorcase. Unfortunately there will be no progress from my side in the next few weeks.

Cerberuser:
Hope this issue isn't stale in fact.

orinokai:
Eagerly awaiting this PR, thanks for everyone's efforts so far. Definitely not stale, issue persists.

Remzi1993:
I'm also waiting for this. @remy Could you check if all is okay?
It seems like it passed all checks.

pepe79:
Hi @remy , I am sorry for you. That's totally understandable and renaming shouldn't be a problem. Sorry that I did not have the time to create a testcase that can be automated. I had some ideas to test this inside docker containers, but hadn't any time to try that...
I will rename the variables and push the changes within the next two hours.

pepe79:
I also created a git repo with a testcase to reproduce this issue and to test this fix (see README): https://github.com/pepe79/test-nodemon-issue-1476

remy:
Picking this up this week - thank you @pepe79 for the changes.

remy:

remy:
I'm in the process or looking at #1633 in detail because this change isn't working as expected in Ubuntu with a standard setup.

I _think_ the problem is when there's more than one sub process, the `kill -0 1234 6789` runs and it gives an error of `code: 1` indicating that one or more of the processes are stopped, assuming both have stopped, the maths is then wrong, `subProcesses.length - returnCode # 2 - 1`.

The logic should either run this function for each sub process or it should revalidate whether the processes are running with pstree. I need to look back through the comments here as to why pstree wasn't used though.

pepe79:
@remy , sorry didn't see that coming. On my system `kill -0` returns exit code 2 if both given pids terminated...
```
> kill -0 4567 6789
kill: kill 4567 failed: no such process
kill: kill 6789 failed: no such process
> echo $?
2
```
Using `pstree` is the much better cross-platform and portable way. Sorry for the inconvience.

Issue #1583, opened 2019-06-26
Option -e seems to be ignored


realtebo:
- `nodemon -v`: 1.19.1
- `node -v`: v10.16.0
- Operating system/terminal environment:  Debian 9.9 - bash
- Using Docker? What image:  NO
- Command you ran:

        nodemon -e js,json --exec "codeceptjs run"

### Expected behaviour

I expected that nodemon will watch for any file in the current directory whis extensions `.js`, or `.json`

### Actual behaviour

When nodemon start, It is watching for _any_ file.

Note that rows stating `watching: *.*`

```
$ nodemon -e js,json --exec "codeceptjs run"
[nodemon] 1.19.1
[nodemon] to restart at any time, enter `rs`
[nodemon] watching: *.*
[nodemon] starting `codeceptjs run`
```

Also, note this relevant portion of the dump
I can see my ext option setup correctly, watch monitor array tell me it's ignorin watching extensions

```
     execOptions:
      { script: null,
        exec: 'codeceptjs run',
        args: [],
        scriptPosition: null,
        nodeArgs: undefined,
        execArgs: [],
        ext: 'js,json',
        env: {} },
     monitor:
      [ '*.*',
        '!**/.git/**',
        '!**/.nyc_output/**',
        '!**/.sass-cache/**',
        '!**/bower_components/**',
        '!**/coverage/**',
        '!**/node_modules/**' ] },

```

realtebo:
Problem is still here

remy:
Wait…have you tested that this _doesn't_ restart on a `.json` file change?

remy:
Closing as this isn't an issue. The text refers to the directories being watch - though I'm pushing a PR that will clarify that language 👍 thanks!

PR #1591, opened 2019-07-17
fix: language around "watching"

remy:
ref: #1583

It was unclear whether it was listing files being watched when in fact
it was the list of directories (plus, make it cleaner)


remy:
Hmm, @remy why didn't you merge this yet? There was a reason I'm sure…

remy:
Why didn't I merge this???

Issue #1604, opened 2019-09-05
Be able to set `option.restartable`


ryusaka:
### Expected behaviour
I'm using some processes of nodemon at the same time. I have to type 'rs' so many times to restart server I want to restart.
So I want a new option `--restartable`.

I think it's easy to add this feature.
just add
```javascript
if (arg === '--restartable') {
  options.restartable = eatNext();
}
```
to
https://github.com/remy/nodemon/blob/12b66cd492b4b06ffc91da57eca4794c25f05632/lib/cli/parse.js

remy:
I'm not sure this is what you want. `restartable` is simply the text you enter to force restart nodemon's subprocess. The default is `rs`, but your new PR will expose this as an option on the command line - which is fine, but it just means you can change it to something else, like typing 'restart' instead of 'rs'.

ryusaka:
In my multi nodemon environment, if I type 'rs', sometimes the other process which is different from what I want to restart is restarted.

I want to set two different command to restart each process.

PR #1605, opened 2019-09-05
feat: add option '--restartable'

ryusaka:
See  #1604 
I know the doc should be added but I'm not native English speaker.
Please someone add that 😂.


remy:
Is restartable the 'rs' thing? (I know I wrote it, but it was years ago and I've not needed to change it!)

ryusaka:
Yes. Now I have to save any file change like a new line to reload the process. 
In my multi nodemon environment, if I use 'rs', sometimes the other process is restarted.

PR #1622, opened 2019-10-22
Update chokidar to v3

radum:
Massive CPU & RAM consumption [improvements](https://github.com/paulmillr/chokidar/releases/tag/3.0.0). 17x package & deps size reduction.

Also this removes support for old Node version, minimum being 8 now. So a major bump will be required also.

With ~2m weekly instals I think nodemon should also benefit from chokidar improvements in terms of size and speed.

### DIFF
--- a/package.json
+++ b/package.json
@@ -9,7 +9,7 @@
     "nodemon": "./bin/nodemon.js"
   },
   "engines": {
-    "node": ">=4"
+    "node": ">=8"
   },
   "repository": {
     "type": "git",
@@ -55,7 +55,7 @@
     "should": "~4.0.0"
   },
   "dependencies": {
-    "chokidar": "^2.1.8",
+    "chokidar": "^3.2.2",
     "debug": "^3.2.6",
     "ignore-by-default": "^1.0.1",
     "minimatch": "^3.0.4",
--- a/package.json
+++ b/package.json
@@ -9,7 +9,7 @@
     "nodemon": "./bin/nodemon.js"
   },
   "engines": {
-    "node": ">=8"
+    "node": ">=8.10.0"
   },
   "repository": {
     "type": "git",
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,11 +5,10 @@ cache:
 notifications:
   email: false
 node_js:
+  - '12'
   - '11'
   - '10'
   - '8'
-  - '6'
-  - '4'
 before_install:
   - if [ "$TRAVIS_PULL_REQUEST_BRANCH" == "" ]; then echo "//registry.npmjs.org/:_authToken=\${NPM_TOKEN}" >> .npmrc; fi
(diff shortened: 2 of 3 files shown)


remy:
IIRC this is also a breaking change as it will drop support for node@4 which nodemon still supports (not actively, but by virtue that the tests pass).

I've been considering this, but wary of a major jump. I need to review in detail exactly what the impact is. (Also to jump in front of those to say "only support LTS" - that's not the aim of the project, nodemon supports "legacy" users who don't have the comfort of being able to follow LTS).

radum:
Yeah I know, I am using nodemon on projects I can't upgrade node on for various reasons. But I think it is sensible and a well used standard to bump the major and instruct everyone that to use it with older versions of node they should stay on the previous version.

The benefits IMHO out-weight the annoyance that could cause for a short period of time until users get the point.

I'm not saying to go LTS only but at least to make sure it uses less resources if it can. I know from experience that it makes a difference, I am actually using a fork with the change just because of the benefits.

remy:
I know v3 drops node 4 (sensibly) but I think it still supports a wide range (beyond LTS) - I just need to review it myself to be sure of what to do and how it impacts. Definitely a "technical" breaking change which will trigger nodemon to v2.x.

package.json
```diff
@@ -9,7 +9,7 @@
     "nodemon": "./bin/nodemon.js"
   },
   "engines": {
-    "node": ">=4"
+    "node": ">=8"
```
XhmikosR: FYI this should be at least 8.10.0

XhmikosR:
chokidar 3.x does not work on anything less than 8.10.0. That being said this is definitely a huge improvement which should come with a major version bump as you already pointed out.

sodatea:
https://github.com/fsevents/fsevents/issues/291

One of the major changes in chokidar v3 is the upgrade of the `fsevents` package. It no longer downloads binaries from a shady AWS service, so we can expect better compatibilities with future Node.js versions.

pipobscure:
While I understand wanting to provide backward compatibility, at this point node versions below v8 should be considered a security risk. So one could argue that compatibility with such old node versions is not doing anyone a favor, but rather a disservice.

On the other hand, we are trying to actively deprecate fsevents v1, because it opens installs to all kinds of issues due to the installation of binaries from AWS. In addition whenever new node-version comes out, new binaries have to be produced, which require a patch version bump every time. This means that `package-lock.json` becomes an active detriment when upgrading node and always causes users issues.

So it really also becomes a question of **supporting old node-version which are a security risk at the cost of causing issues for new node-versions**. And from that perspective it should be a clear choice.

If someone asked you: **"Do you want to support node v6 or nodev13, you have to pick one over the other"** then the answer should be clear. And that is in fact the question being asked.

remy:

Issue #1626, opened 2019-11-14
Nodemon boot failure


pipeng119:
- `nodemon -v`:1.19.4
- `node -v`:12.10.0 | 11.15.0
- Operating system/terminal environment: MacOS
- Using Docker? What image: no
- Command you ran: nodemon server.js

### Expected behaviour
Simply run a node services,Only the following code in the service
```
const path = require('path');
console.log(1)
```

### Actual behaviour

Command-line error:`[nodemon] Internal watch failed: EMFILE: too many open files, watch`
### Steps to reproduce
`nodemon server.js`

---

I tried the following solutions：
`ulimit -n`
Used to increase the file open upper limit，But there is no effect

remy:
Pretty sure you're watching way too many files - potentially in the root directory? Maybe move deeper into your project.

To test this you can use the debug mode to show how many files are being watched:

```
DEBUG=nodemon:* nodemon
```

I've also pushed a change that catches this error and gives better suggestions.

remy:
I've pushed the PR live in nodemon@2.0.0 - to notify and offer better advice when that particular error occurs.

PR #1631, opened 2019-11-20
 fix: make error on too many files clear

remy:
Ref #1626


remy:

Issue #1638, opened 2019-11-24
debug.js colour and logging behaviour changed with the upgrade to v2


noway:
- `nodemon -v`: 2.0.1
- `node -v`:v12.6.0
- Operating system/terminal environment: Mac OS Catalina 10.15.1, iTerm 2 3.3.7
- Using Docker? What image: no
- Command you ran: nodemon

### Expected behaviour
![image](https://user-images.githubusercontent.com/2031472/69502624-1882bc00-0f76-11ea-93d9-809e6453cab6.png)

### Actual behaviour
![image](https://user-images.githubusercontent.com/2031472/69502627-28020500-0f76-11ea-9782-152e2a1cfdc0.png)

### Steps to reproduce
use debug.js on v1 and v2. on v1 logging suffixes and colour shown on all lines, on v2 logging suffixes shown only on first colour with no colour.

---

If applicable, please append the `--dump` flag on your command and include the output here **ensuring to remove any sensitive/personal details or tokens**.

remy:
I think I know why this is, but can you provide a sample repo (reduced down as much as possible) so that I can replicate against?

noway:
thank you for response, sorry don't have enough bandwidth for that 😥

remy:
Well, same here, but can you at least share some part of your code so that someone else has a chance of replicating what colour lib you're using?

noway:
made a repro repo here: https://github.com/noway/debugjsnodemonbug

debug.js colours stuff itself by the looks of it

gilles-crealp:
Same problem here, since 2.0.1 (**not in 2.0.0**). The problem is with the package [`supportsColor`](https://github.com/chalk/supports-color). It now return `false` with `supportsColor.stderr` (but not with `supportsColor.stdout`).

gilles-crealp:
> I think I know why this is, but can you provide a sample repo (reduced down as much as possible) so that I can replicate against?

```javascript
const supportsColor = require('supports-color')
if (supportsColor.stdout) {
  console.log('Terminal stdout supports color')
} else {
  console.log('Terminal stdout DO NOT supports color')
}
if (supportsColor.stderr) {
  console.log('Terminal stderr supports color')
} else {
  console.log('Terminal stderr DO NOT supports color')
}
```

**With nodemon@2.0.0**
```
Terminal stdout supports color
Terminal stderr supports color
```

**With nodemon@2.0.1**
```
Terminal stdout supports color
Terminal stderr DO NOT supports color
```

[...]

PR #1661, opened 2020-01-14
fix: fixes issue #1476 after introduced again by version 2.0.2

marcelkottmann:
Hey @remy , seems like issue #1476 was introduced again by version 2.0.2. Please look at my new comment in https://github.com/remy/nodemon/issues/1476#issuecomment-573592914

We have to wait for all subprocess to shutdown before killing the main process. Killing the main process has the side effect of emitting the exit-event, which in turn starts the program again (too early) before the child processes have been terminated.

I updated the test repository to reproduce this error with nodemon@2.0.2 and to test the fix: https://github.com/marcelkottmann/test-nodemon-issue-1476

### DIFF
--- a/lib/monitor/run.js
+++ b/lib/monitor/run.js
@@ -321,7 +321,7 @@ function run(options) {
 function waitForSubProcesses(pid, callback) {
   debug('checking ps tree for pids of ' + pid);
   psTree(pid, (err, pids) => {
-    if (!pids.length) {
+    if (err || !pids.length) {
       return callback();
     }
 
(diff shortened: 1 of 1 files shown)


hphvac:
Would be nice, this problem still exists for me

remy:
The change in the order which the processes are stopped was intentional. Specifically this change: https://github.com/remy/nodemon/commit/47dfb8b

Can you check that your change still allows the origin issue to be solved? https://github.com/remy/nodemon/issues/1638 (I appreciate this was from a refactor with the release of 2.0.1 that simplified your original code - which was also fixed bugs on linux platforms)

hphvac:
I am currently running 2.0.2. I believe that the issue you referenced doesn't apply to me.

PR #1666, opened 2020-01-24
making website a bit stunning

darkCavalier11:
-Hovering the logo create a nice effect.
-centering texts
-adding favicon and others

### DIFF
--- a/index.html
+++ b/index.html
@@ -1,0 +1,218 @@
+<!DOCTYPE html>
+<html>
+<head>
+<meta charset=utf-8>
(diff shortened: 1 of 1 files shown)


darkCavalier11:
pls review the code

remy:
I have. Sorry, won't merge. The changes were more subjective design changes rather than being for better UX, and in fact the underlines were removed from links is an example of where it wasn't clear what was a link and what wasn't (most prominent on the footer). Thank you all the same.

PR #1669, opened 2020-01-28
docs: Use HTTPS protocol for homepage link

jjasonclark:
docs: use HTTPS protocol for homepage link

update to project's `package.json` and `README.md` sample `package.json` `homepage` value

### DIFF
--- a/README.md
+++ b/README.md
@@ -108,7 +108,7 @@ Specify the config in the same format as you would for a config file but under `
 ```json
 {
   "name": "nodemon",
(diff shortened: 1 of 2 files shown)


jjasonclark:
What does the `ci/semantic-release` check do? Happy to update.

remy:
it checks and validates whether the commit message format is correct as per: https://github.com/remy/nodemon/blob/master/.github/CONTRIBUTING.md#commit-messages

You'll want to change it to be prefixed with `docs: ` IIRC.

Issue #1742, opened 2020-07-23
--on-change-only / -C seemingly does nothing


ejulen:
- `nodemon -v`: 2.0.4
- `node -v`: v12.17.0
- Operating system/terminal environment: Ubuntu 20.04
- Command you ran: `nodemon -C some-script.js`

### Expected behaviour
some-script.js not being executed on startup. Please correct me if this is a misunderstanding on my part!

### Actual behaviour
some-script.js is executed on startup.

### Steps to reproduce
1. Install `nodemon`.
2. Create any sort of script, for example something that simply logs "Hello!" to the console.
3. Run `nodemon -C script-name-here.js`.

---

<details>
<summary>Details</summary>

```
[nodemon] 2.0.4
[nodemon] to restart at any time, enter `rs`
[nodemon] watching path(s): *.*
[nodemon] watching extensions: js,mjs,json
--------------
node: v12.17.0
nodemon: 2.0.4
command: /home/johannes/.local/bin/node /home/johannes/dev/project/node_modules/nodemon/bin/nodemon.js -C --dump some-script.js
cwd: /home/johannes/dev/project
OS: linux x64
--------------
{
  run: false,
  system: { cwd: '/home/johannes/dev/project' },
  required: false,
  dirs: [ '/home/johannes/dev/project' ],
  timeout: 1000,
  options: {
    runOnChangeOnly: true,
    dump: true,
    ignore: [
      '**/.git/**',
      '**/.nyc_output/**',
      '**/.sass-cache/**',
      '**/bower_components/**',
      '**/coverage/**',
      '**/node_modules/**',
      re: /.*.*\/\.git\/.*.*|.*.*\/\.nyc_output\/.*.*|.*.*\/\.sass\-cache\/.*.*|.*.*\/bower_components\/.*.*|.*.*\/coverage\/.*.*|.*.*\/node_modules\/.*.*/
    ],
    watch: [ '*.*', re: /.*\..*/ ],
    monitor: [
      '*.*',
      '!**/.git/**',
      '!**/.nyc_output/**',
      '!**/.sass-cache/**',
      '!**/bower_components/**',
      '!**/coverage/**',
      '!**/node_modules/**'
    ],
    ignoreRoot: [
      '**/.git/**',
      '**/.nyc_output/**',
      '**/.sass-cache/**',
      '**/bower_components/**',
      '**/coverage/**',
      '**/node_modules/**'
    ],
    restartable: 'rs',
    colours: true,
[...]

PR #1747, opened 2020-08-04
fix for issue #1742: corrected run.js, run.test.js, watch.js for the use case runOnChangeOnly=true

MonarchChakri:
Updated

1. lib/monitor/run.js: to handle the start on change only case (-C/--on-change-only) by adding an early exit.
2. test/monitor/run.test.js: to run the test which was previously skipped.
3. lib/monitor/watch.js: to precheck config.options.execOptions before accessing  script.


MonarchChakri:
Closing to make resubmit and pass semantic check

PR #1750, opened 2020-08-06
chore: update dependencies

GChuf:
Fixes 10 low vulnerabilities , updates dependencies.

### DIFF
--- a/package.json
+++ b/package.json
@@ -42,29 +42,29 @@
     "postinstall": "node bin/postinstall || exit 0"
   },
   "devDependencies": {
-    "@commitlint/cli": "^3.1.3",
+    "@commitlint/cli": "^3.2.0",
     "@commitlint/config-angular": "^3.1.1",
-    "async": "1.4.2",
-    "coffee-script": "~1.7.1",
+    "async": "^1.5.2",
+    "coffeescript": "^2.5.1",
     "husky": "^0.14.3",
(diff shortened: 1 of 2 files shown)


remy:
Does this fix production vulns - because I didn't think nodemon was carrying any.

If I'm doing local dev, then I think some of the packages are marked as potentially vulnerable but that seems to only really affect me…

GChuf:
Yeah, that's only for local dev I guess. But hopefully the performance of nodemon increases as well, I haven't tested that ...

package.json
```diff
@@ -42,10 +42,10 @@
     "postinstall": "node bin/postinstall || exit 0"
   },
   "devDependencies": {
-    "@commitlint/cli": "^3.1.3",
+    "@commitlint/cli": "^3.2.0",
     "@commitlint/config-angular": "^3.1.1",
-    "async": "1.4.2",
-    "coffee-script": "~1.7.1",
+    "async": "^1.5.2",
+    "coffee-script": "~1.12.7",
```
GeoffreyBooth, line 48: Please change this to `"coffeescript"`, the project migrated.

GChuf:
oops - updated just the package.json file
will fix it so that package-lock gets updated as well later today

GChuf:
OK - squashed commits and updated coffee-script to coffeescript. Tests passed.

PR #1751, opened 2020-08-07
fix: runOnChangeOnly=true

MonarchChakri:
Updated

1. test/monitor/run.test.js: to run the test which was previously skipped.
2. lib/monitor/run.js: to handle the start on change only case (-C/--on-change-only) by adding an early exit.
3. lib/monitor/watch.js: to precheck config.options.execOptions before accessing script.

### DIFF
--- a/lib/monitor/run.js
+++ b/lib/monitor/run.js
@@ -18,15 +18,30 @@ var signals = require('./signals');
 
 function run(options) {
   var cmd = config.command.raw;
+  // moved up
+  // we need restart function below in the global scope for run.kill
+  /*jshint validthis:true*/
+  restart = run.bind(this, options);
+  run.restart = restart;
+
+  // binding options with instance of run
+  // so that we can use it in run.kill
+  run.options = options;
 
   var runCmd = !options.runOnChangeOnly || config.lastStarted !== 0;
   if (runCmd) {
     utils.log.status('starting `' + config.command.string + '`');
+  } else {
+    // should just watch file if command is not to be run
+    // had another alternate approach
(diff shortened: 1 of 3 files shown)


MonarchChakri:
@remy Did you get time to look at this PR?

remy:
Caught me at a personal time that I was actively avoiding the computer. I must have looked at it once, but I couldn't see tests for such significant changes. Happy to reopen if there's checks for the specific changes (that either shows fixes have been applied or otherwise).

MonarchChakri:
@remy
This commit was given to fix the issue https://github.com/remy/nodemon/issues/1742: --on-change-only / -C options were not working and it was marked as a bug.
Found the root cause of the issue and fixed it in this commit.

Updated the test case to be run now in **`test/monitor/run.test.js`** to run the test which was previously skipped.
Also it was not caught by the tests because the test was never actually executed under test/monitor/run.test.js.
Please review and merge.

image

MonarchChakri:
Please verify the test case for runOnChangeOnly under **`test/monitor/run.test.js`**

remy:
Okay, I'm always super wary of large code changes, even if they're rearranging, but this change still has all tests passing so either the 100s of tests I've written are garbage or your code hasn't broken anything 👍 

Merging now.

MonarchChakri:
Thanks for merging and adding the label **`hacktoberfest-accepted`**. 😄 
But as per the rules I need to submit a PR during October and I have submitted it long back.
Maybe assign me another issue which I can work on it or I'll try next time. 😅

PR #1765, opened 2020-09-17
fix: respect env.OPENCOLLECTIVE_HIDE

woodcockjosh:
Respect official environment variable to hide codecollective notifications OPENCOLLECTIVE_HIDE

### DIFF
--- a/bin/postinstall.js
+++ b/bin/postinstall.js
@@ -1,7 +1,7 @@
 #!/usr/bin/env node
 
 function main() {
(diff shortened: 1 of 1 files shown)


XhmikosR:
@remy have you thought about hiding the postinstall message on CI, if `process.env.CI` is truthy?

remy:
@XhmikosR happy to add that - but you really shouldn't be installing nodemon in your CI env! (ie. only prod deps)

Issue #1779, opened 2020-10-12
Entrypoint script is filtered out, preventing restart


nprbst:
```
→ nodemon -v
2.0.4

→ node -v
v12.18.4

→ uname -mprs
Darwin 19.6.0 x86_64 i386

→ cat nodemon.json 
{
  "verbose": true,
  "watch": ["index"],
  "ext": "ts,js,json",
  "ignore": ["*.test.ts"],
  "delay": "1",
  "env": {
    "NODE_ENV": "development"
  },
  "execMap": {
    "ts": "ts-node"
  }
}

→ cat index.ts 
const main = () => {
  console.log("Hello World");
};
main();

→ nodemon
...
Hello World!
[nodemon] clean exit - waiting for changes before restart
[nodemon] files triggering change check: index.ts
[nodemon] changes after filters (before/after): 1/0
```

### Expected behaviour

Changes to the entrypoint script should trigger restart.

### Actual behaviour

Changes to the entrypoint script do not trigger restart. Entrypoint script is filtered out due to a bug in `watch.js`. PR coming shortly.

### Steps to reproduce

See above terminal history.

remy:
Um, this actually correct. There's no match on the rules - because the nodemon config says to restart on a file called `index`, but your file is `index.js`.

The `watch` is either for filenames (like `index.js`) or for directories (like `./src/*`).

To summarise: this isn't a bug.

PR #1780, opened 2020-10-12
Correct entrypoint script path comparison bug

nprbst:
The previously existing logic incorrectly compares an absolute path with a relative path, preventing a change in the entrypoint script from triggering a restart. This PR correctly compares paths relative to cwd, as `files` are represented.


nprbst:
Fixes #1779

remy:
This needs a test to show the previous logic was wrong and that this change will fix it.

I know the change is only two lines, but they don't look trivial (I'm wondering if the second change was originally related to windows - I can't honestly remember).

remy:
Based on the original filed issue - I don't believe this was an issue in the first place. Happy to reopen if there's either a test that shows a bug or if there's a misunderstanding in the original issue.

Issue #1343, opened 2018-05-13
`npm audit` vulnerability issues


ehmicky:
- `nodemon -v`: `master: ff79835b143290a53d495d0f4263e5ef45df2c51`
- `node -v`: `v10.1.0`
- Operating system/terminal environment: Ubuntu 18.04 
- Command you ran: `npm audit`

### Expected behaviour

It should print no warnings.

### Actual behaviour

It prints:
```
                                                                                
                       === npm audit security report ===                        
                                                                                
# Run  npm install --dev mocha@5.1.1  to resolve 4 vulnerabilities
SEMVER WARNING: Recommended action is a potentially breaking change
┌───────────────┬──────────────────────────────────────────────────────────────┐
│ Low           │ Regular Expression Denial of Service                         │
├───────────────┼──────────────────────────────────────────────────────────────┤
│ Package       │ debug                                                        │
├───────────────┼──────────────────────────────────────────────────────────────┤
│ Dependency of │ mocha [dev]                                                  │
├───────────────┼──────────────────────────────────────────────────────────────┤
│ Path          │ mocha > debug                                                │
├───────────────┼──────────────────────────────────────────────────────────────┤
│ More info     │ https://nodesecurity.io/advisories/534                       │
└───────────────┴──────────────────────────────────────────────────────────────┘


┌───────────────┬──────────────────────────────────────────────────────────────┐
│ Moderate      │ Regular Expression Denial of Service                         │
├───────────────┼──────────────────────────────────────────────────────────────┤
│ Package       │ ms                                                           │
├───────────────┼──────────────────────────────────────────────────────────────┤
│ Dependency of │ mocha [dev]                                                  │
├───────────────┼──────────────────────────────────────────────────────────────┤
│ Path          │ mocha > debug > ms                                           │
├───────────────┼──────────────────────────────────────────────────────────────┤
│ More info     │ https://nodesecurity.io/advisories/46                        │
└───────────────┴──────────────────────────────────────────────────────────────┘


┌───────────────┬──────────────────────────────────────────────────────────────┐
│ High          │ Regular Expression Denial of Service                         │
├───────────────┼──────────────────────────────────────────────────────────────┤
│ Package       │ minimatch                                                    │
├───────────────┼──────────────────────────────────────────────────────────────┤
│ Dependency of │ mocha [dev]                                                  │
├───────────────┼──────────────────────────────────────────────────────────────┤
│ Path          │ mocha > glob > minimatch                                     │
[...]

PR #1785, opened 2020-10-18
Switching from JSCS to ESLint

Antrikshy:
JSCS reached EOL 4 years ago. I noticed that there has been chatter about switching to ESLint in various issues, including #983 and #1343. As a bonus, this also resolves a handful of dev-deps security vulnerabilities as reported by `npm audit`.

Surprisingly, there were no issues doing this, and it didn't require touching a lot of files. On the other hand, I guess this isn't too surprising. I used Polyjuice [as recommended by](https://eslint.org/docs/user-guide/migrating-from-jscs) the ESLint docs to migrate an existing JSCS configuration. It generated everything inside the `rules` object in the new .eslintrc.json.

For good measure, I also went through the [ESLint `--init` flow](https://github.com/eslint/eslint#installation-and-usage) to generate the rest of the config, and merged the two. I recommend going through it yourself while reviewing this change. I'm open to using different settings. Simply run:

```
npm install --save-dev eslint
npx eslint --init
```

I answered the questionnaire as follows:

```
$~/Projects/nodemon$ npx eslint --init
✔ How would you like to use ESLint? · syntax
✔ What type of modules does your project use? · commonjs
✔ Which framework does your project use? · none
✔ Does your project use TypeScript? · No / Yes
✔ Where does your code run? · browser
✔ What format do you want your config file to be in? · JSON
Successfully created .eslintrc.json file in /home/antriksh/Projects/nodemon
```

**Small note:** The `commitmsg` npm script seems to throw a strange error. I used `--no-verify` while committing to bypass whatever it does (I presume just verifies my commit message, which I hope is satisfactory).


remy:
This PR is good (and yeah, I noticed something similar with the commit check last week too, but your commit message looks good). The only small-ish concern is that node 8 isn't supported by eslint (or at least uses syntax that doesn't work according to travis). 

The bigger step is to drop node 8 support from nodemon (which is a sensible move either way, but requires some careful thinking on my part as to whether it's a breaking change or not).

An intermediate step is to stop testing node 8 in Travis to let this merge…

I think the later is probably the right way to go. Just let me have a few hours to think about it.

remy:
Okay, I'm going to drop node 8 in travis. Your PR also lets me (finally) upgrade the source code and move to prettier configs that I've been using for years now and maybe make some changes and clean ups in the source code that I've been wanting to do but didn't have the motivation.

PR #1790, opened 2020-10-26
Update run.js to include exit code in crash log

EtzBetz:
In my opinion the crash log should include the exit code

### DIFF
--- a/lib/monitor/run.js
+++ b/lib/monitor/run.js
@@ -239,8 +239,8 @@ function run(options) {
           process.exit(1);
         }
       } else {
-        utils.log.fail('app crashed - waiting for file changes before' +
-          ' starting...');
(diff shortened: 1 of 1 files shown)


remy:
Why? It's great you've got an opinion but right now only you know what that is...

EtzBetz:
Well when my application closes unexpectedly, I want to know why it closed.

remy:
The exit code won't give you that. You'll see `1` the case that your program exited "unexpectedly" (rather than nodemon restarting due to change). There's no error exit code you'll see…

EtzBetz:
I thought that `code` was the code that node exited with? Isn't there a way to get the exit code of node?

remy:
`1` is the exit code from node...

EtzBetz:
Not when I exit in node code myself

PR #1797, opened 2020-11-16
fix: typo: intial -> initial

josephfrazier:


remy:
thanks, but I'll include this typo change in another release later or (or more likely: remove the comments entirely).

Rather than deploy a new version of nodemon for a comment change.

PR #1821, opened 2021-01-04
Update package.json

daRasmussen:
Update package.json


remy:
Explain?

daRasmussen:
Hello/ 

I hade problems with `npm install nodemon --save-dev` on a windows version 1909 build 18363.1256 machine running powershell. The script can't complete due to the ` || ` operation in postinstall. If I download the code and run install `npm install ../nodemon-master` with this change install works. But other packages that use nodemon as a dependency won't work.  

I tried the same operation `npm install nodemon --save-dev` later on another windows machine version 20xx and later build, and it worked. 

So I think this might be an windows specific issue? But I am not sure, if possible it would be great if the install script could be run on older windows builds, but I am not sure if this is possible to accomplish without removing `||` operation.

PR #1839, opened 2021-02-01
added favicon to web application

vikramadityasinghs:


remy:
There's an svg and png in the website root - so no need to link to another domain for this.

Issue #1720, opened 2020-05-14
Nodemon is not able to gracefully restart applications in Windows environment


chriswoodle:
Possibly related to https://github.com/remy/nodemon/issues/1567

- `nodemon -v`: 6.13.4
- `node -v`: v12.16.1
- Operating system/terminal environment: 
    - Windows Command Prompt, Windows Git Bash, Windows Terminal WSL
    ```
         Edition: Windows 10 Pro 
         Version: 1909
    Installed on: 5/22/2019
        OS Build: 18363.836 
    ```

- Command ran: `nodemon --signal SIGHUP index.js`

### Expected behaviour

On Windows, Nodemon should wait for the process to complete if it handles a shut down signal. 

> Notice the `express server closed` message 

WSL:
![wsl](https://github.com/chriswoodle/nodemon-1567/raw/master/wsl.png)

### Actual behaviour

The application is killed and restarted right away. This behavior does not exist in Linux, Mac and WSL. 

CMD:
![windows](https://raw.githubusercontent.com/chriswoodle/nodemon-1567/master/windows.png)

### Steps to reproduce

I have created a sample Express app here: https://github.com/chriswoodle/nodemon-1567 that shuts down the express server once the application is signaled to stop `['SIGTERM', 'SIGINT', 'SIGHUP']`

```
git clone https://github.com/chriswoodle/nodemon-1567.git
cd nodemon-1567
npm install
npm start
touch index.js
```

Let me know if I can provide any more information or test changes, thanks.

---

remy:
If I get time, I'll look into this, but I wanted to stop and say:

Thank you for filing an issue with a good level of detail and a good example of pared down code to replicate the issue. I'll be pointing people to this issue as an example of what makes the world of difference in debugging. 🙏

nicholas-ochoa:
I came here with the same exact problem and I'm glad I found this issue. On *nix this works perfectly, on Windows it's almost like the signals are ignored and a SIGKILL is used instead? I'm seeing the same behavior regardless during restarts for file changes (or typing 'rs') - the application is IMMEDIATELY restarted and my shutdown handlers are never called.

JulianNicholls:
Hi Remy.

Just a suggestion

I hadn't seen this problem on Windows via WSL until I enabled a subscription in my GraphQL server. 
GraphQL Subscriptions use websockets, which may be relevant.

countzero:
Hi there,

[...]

PR #1853, opened 2021-04-29
fix: add support for SIGINT on Windows (fixes issue #1720)

countzero:
This adds a `windows-kill.exe` binary which handles the sending of the POSIX signal `SIGINT` on Windows.

The repository https://github.com/countzero/nodemon_windows_kill contains a a cross-platform compatible test to quickly verify the correct behaviour of this fix.


remy:
I've published a debug build to npm - `npm i -g nodemon@debug` should give you `2.0.8-alpha.a` - let's get some windows testing with that and ship this PR.

Really great work seeing this through 👍

countzero:
@remy: I added the a hard shutdown behaviour on `SIGKILL` only. This supports the use case of @wrexbe and is a nice addition!

chriswoodle:
`2.0.8-alpha.a` works fine for me on windows.

remy:
Right, this has been quiet for a bit, but let's ship this - I reckon it'll help those windows users - I've seen quite a few similar issues recently so hopefully this nails it.

Really excellent work @countzero

remy:
FFS - something has gone weird on travis - so this isn't actually live, so I'll need to fix whatever has happened _then_ this will be live.

RepentAndBelieveTheGospel:
I get this with the latest version.

image

The system cannot find the file C:\Projects\node.js\complete-node-bootcamp\4-natours\starter\node_modules\nodemon\bin\windows-kill.exe.

I've initially installed nodemon globally, without adding it to the project dependencies. I've fixed this by doing a npm i nodemon --save-dev too.

remy:
@RepentAndBelieveTheGospel issue being tracked here: https://github.com/remy/nodemon/issues/1872

RepentAndBelieveTheGospel:
@remy Sorry, didn't check the opened issues.

bentinata:
I'm not well-versed on Windows system and binaries, and I know compiling it is too much hassle for users. But can we have the source to the executables?

hrieke:
@bentinata I believe the source code is from:
https://github.com/ElyDotDev/node-windows-kill
MIT Licensed

PR #1927, opened 2021-10-03
Update README.md

gautamabhishek1:
DOC: Added a dev script for package.json files that can be used during development.


remy:
Thank you but not sure it's needed. This fall's more under npm docs rather than nodemon.

PR #1967, opened 2022-01-06
docs: added restart event bypass example

notyetspecified:


remy:
Thank you but can you add this to the FAQ and make it into a question. I imagine it's useful, but niché enough that it doesn't belong in the main readme.

Issue #1951, opened 2021-11-08
'ts-node' is not recognized as an internal or external command


milon27:
> nodemon ./src/index.ts

[nodemon] 2.0.14
[nodemon] to restart at any time, enter `rs`
[nodemon] watching path(s): *.*
[nodemon] watching extensions: ts,json
[nodemon] starting `ts-node ./src/index.ts`
'ts-node' is not recognized as an internal or external command,
operable program or batch file.
[nodemon] app crashed - waiting for file changes before starting...

remy:
And what happens when you run the `ts-node` command by itself without nodemon?

TheLordDrake:
Same issue here. 

Package.json:
```
"scripts": {
    "dev": "nodemon main.ts"
},
"devDependencies": {
    "nodemon": "^2.0.14",
    "ts-node": "^10.4.0",
    "typescript": "^4.4.4"
}
```

Calling Nodemon:
```
$ npm run dev

> personalamt-server@0.1.0 dev
> nodemon main.ts

[nodemon] 2.0.14
[nodemon] to restart at any time, enter `rs`
[nodemon] watching path(s): *.*
[nodemon] watching extensions: ts,json
[nodemon] starting `ts-node main.ts`
'ts-node' is not recognized as an internal or external command,
operable program or batch file.
[nodemon] app crashed - waiting for file changes before starting...
```
Calling ts-node:
```
$ npx ts-node main.ts
Hello World
```

milon27:
it was working fine with npm 6.*.* but whenever i have updated it to 8.*.* it no more working. maybe nodemon need to be updated somewhere to work with npm v8.

hope this will fix soon.

milon27:
> 
> 
> Same issue here.
> 
> Package.json:
> 
> ```
> "scripts": {
>     "dev": "nodemon main.ts"
> },
> "devDependencies": {
>     "nodemon": "^2.0.14",
>     "ts-node": "^10.4.0",
>     "typescript": "^4.4.4"
> }
> ```
> 
> Calling Nodemon:
> 
> ```
> $ npm run dev
> 
> > personalamt-server@0.1.0 dev
> > nodemon main.ts
> 
> [nodemon] 2.0.14
> [nodemon] to restart at any time, enter `rs`
> [nodemon] watching path(s): *.*
> [nodemon] watching extensions: ts,json
> [nodemon] starting `ts-node main.ts`
> 'ts-node' is not recognized as an internal or external command,
[...]

PR #1969, opened 2022-01-08
fix: use platform-specific separator when augmenting PATH

lorenzodallavecchia:
Fixes: #1951

### DIFF
--- a/lib/monitor/run.js
+++ b/lib/monitor/run.js
@@ -64,7 +64,7 @@ function run(options) {
 
   const spawnOptions = {
     env: Object.assign({}, process.env, options.execOptions.env, {
-      PATH: binPath + ':' + process.env.PATH,
+      PATH: binPath + path.delimiter + process.env.PATH,
     }),
     stdio: stdio,
   };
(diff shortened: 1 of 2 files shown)


lorenzodallavecchia:
The `--delay` test is also failing on `main`, with no changes since the last time it passed (odd).
I'm positive that the PR is not the cause.

lorenzodallavecchia:
Apparently, 8 seconds is no longer enough for that test I increased the timeout to 12 seconds. It should work now.

remy:
With random failing tests I tend to just re-run (just not on email much on Sunday). Nodemon has a lot of brittle-like tests that require lots of spinning plates, so these tests sometimes do just bork randomly.

lorenzodallavecchia:
Thanks for attending to this PR @remy.
I know full well the pains of unstable tests...

The failing one is probably not time-related, as it completes in about 1 second on my computer, so a 30s timeout should be more than sufficient.

What should I do now? I don't think I can rerun tests myself.
No hurry: I'd like to know if I can do something.
Thanks.

lorenzodallavecchia:
Superseded by #1981.

stijnherreman:
I didn't notice you submitted a PR long before me, sorry! But your suggested fix has eventually made it into a release :)

lorenzodallavecchia:
@stijnherreman, no worries! :)

PR #1979, opened 2022-02-18
chore: convert most var to const and let

jimmywarting:


remy:
Rather too big of a PR and not really needed - but appreciated all the same (sorry, I should have said originally).

PR #1981, opened 2022-02-23
fix: support windows by using path.delimiter

stijnherreman:
Fixes: #1951

This change was suggested in https://github.com/remy/nodemon/issues/1951#issuecomment-1003362605 and I've verified it fixes the issue.

### DIFF
--- a/lib/monitor/run.js
+++ b/lib/monitor/run.js
@@ -64,7 +64,7 @@ function run(options) {
 
   const spawnOptions = {
     env: Object.assign({}, process.env, options.execOptions.env, {
-      PATH: binPath + ':' + process.env.PATH,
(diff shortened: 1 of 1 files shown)


cwadrupldijjit:
Ping @remy.  This seems to be a sane way to fix the problem and all of the tests pass.  Any chance that this could get merged in and released?

remy:
sorry for delay - "life" getting in the way - thank you for this 👍

spazmodius (commented):
Shouldn't we use OS-sensitive path separators, too?

```suggestion
const binPath = path.join(process.cwd(), 'node_modules', '.bin');
```

stijnherreman:
@spazmodius Yes, that change makes sense. Fortunately forward slashes are accepted on Windows, so the current implementation works too.

Issue #1989, opened 2022-03-14
Nodemon spawns have invalid PATH on Windows


spazmodius:
On Windows, nodemon spawns its child process with a modified PATH that is invalid for Windows.

If `nodemon`'s PATH is (an abbreviated example)

`PATH=C:\Program Files (x86)\Microsoft Visual Studio;C:\Program Files (x86)\Windows Kits\10\bin\x64`

then the spawned command runs with a PATH

`PATH=C:\dev/node_modules/.bin:C:\Program Files (x86)\Microsoft Visual Studio;C:\Program Files (x86)\Windows Kits\10\bin\x64`

Notice that **(cwd) + /node_modules/.bin:** has been prepended to the first PATH directory.

- Versions: node@v16.14.0, win32@10.0.19044
- `nodemon -v`: 2.0.15
- Operating system/terminal environment (powershell, gitshell, etc): Windows `cmd`

### Command you ran: 
```
C:\dev> nodemon build.cmd
```

where `build.cmd` contains:
```
path
```

and `nodemon.json` contains:
```
{
	"ext": "cmd,cpp,h",
	"execMap": {
		"cmd": "cmd /c"
	}
}
```

### Expected behaviour
Spawned environments should inherit the same PATH, or at least a valid one.

### Actual behaviour

Spawned environments inherit a corrupted PATH.

---

```
C:\dev>nodemon --dump build.cmd
[nodemon] 2.0.15
[nodemon] to restart at any time, enter `rs`
[nodemon] watching path(s): *.*
[nodemon] watching extensions: cmd,cpp,h
--------------
node: v16.14.0
nodemon: 2.0.15
command: C:\Portable\node\node.exe C:\dev\node_modules\nodemon\bin\nodemon.js --dump build.cmd
cwd: C:\dev
OS: win32 x64
--------------
{
  run: false,
  system: { cwd: 'C:\\dev' },
  required: false,
  dirs: [ 'C:\\dev' ],
  timeout: 1000,
  options: {
    dump: true,
    execMap: { cmd: 'cmd /c', py: 'python', rb: 'ruby', ts: 'ts-node' },
    quiet: false,
    ignore: [
      '**/.git/**',
      '**/.nyc_output/**',
      '**/.sass-cache/**',
      '**/bower_components/**',
      '**/coverage/**',
      '**/node_modules/**',
[...]

PR #1994, opened 2022-03-25
fix: resolve path is corrupt when spawning on windows (#1989)

Sikkesoft:
This fixes the issue #1989;
When using the spawn-method, the path was prefixed with the cwd() +'/node_modules/.bin' + ':', resulting on Windows in a corrupt first entry. 
This will largely go unnoticed, until one attempts to execute a program located in that folder. 

The fix composes the folder and adds it to the path in a platform dependent way 
The added tests work on Windows native and WSL2-Ubuntu, but may fail when encountering folders containing extended characters such as é, ä and so on. 
Testing if the folder exists proofed unsuccessful; my path contained several folders that didn't  exist.
Alternately, attempting to launch an (any) executable from the first folder in the path is playing Russian roulette.

### DIFF
--- a/lib/monitor/run.js
+++ b/lib/monitor/run.js
@@ -59,12 +59,13 @@ function run(options) {
 
   var sh = 'sh';
   var shFlag = '-c';
+  const pathSep = utils.isWindows?';':':';
 
-  const binPath = process.cwd() + '/node_modules/.bin';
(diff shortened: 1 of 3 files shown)


Sikkesoft:
Because the problem actually is the same as #1951. This Pull-request I am abandoning the pull request. Feel free to include the unit-test

PR #2032, opened 2022-06-23
Update bug_report.md

elrumordelaluz:
Remove backslash which throws an error (tested on OSX fish and CentOS bash).

before
Screenshot 2022-06-23 at 16 58 34

after
Screenshot 2022-06-23 at 16 58 48


remy:
ta - no idea why I put the backslashes in there…

PR #2033, opened 2022-06-24
fix: Replace update notifier with simplified deps

### LINKED ISSUE #1961: Consider removing/replacing update-notifier
This is the dependency tree I got from installing `nodemon` in an empty project:
```
└─┬ nodemon@2.0.15
  ├─┬ chokidar@3.5.2
  │ ├─┬ anymatch@3.1.2
  │ │ ├── normalize-path@3.0.0 deduped
  │ │ └── picomatch@2.3.0
  │ ├─┬ braces@3.0.2
  │ │ └─┬ fill-range@7.0.1
  │ │   └─┬ to-regex-range@5.0.1
  │ │     └── is-number@7.0.0
  │ ├── fsevents@2.3.2
  │ ├─┬ glob-parent@5.1.2
  │ │ └── is-glob@4.0.3 deduped
  │ ├─┬ is-binary-path@2.1.0
  │ │ └── binary-extensions@2.2.0
  │ ├─┬ is-glob@4.0.3
  │ │ └── is-extglob@2.1.1
  │ ├── normalize-path@3.0.0
  │ └─┬ readdirp@3.6.0
  │   └── picomatch@2.3.0 deduped
  ├─┬ debug@3.2.7
  │ └── ms@2.1.3
  ├── ignore-by-default@1.0.1
  ├─┬ minimatch@3.0.4
  │ └─┬ brace-expansion@1.1.11
  │   ├── balanced-match@1.0.2
  │   └── concat-map@0.0.1
  ├── pstree.remy@1.1.8
  ├── semver@5.7.1
  ├─┬ supports-color@5.5.0
  │ └── has-flag@3.0.0
  ├─┬ touch@3.1.0
  │ └─┬ nopt@1.0.10
  │   └── abbrev@1.1.1
  ├── undefsafe@2.0.5
  └─┬ update-notifier@5.1.0
    ├─┬ boxen@5.1.2
    │ ├─┬ ansi-align@3.0.1
    │ │ └── string-width@4.2.3 deduped
    │ ├── camelcase@6.2.1
    │ ├── chalk@4.1.2 deduped
    │ ├── cli-boxes@2.2.1
    │ ├─┬ string-width@4.2.3
    │ │ ├── emoji-regex@8.0.0
    │ │ ├── is-fullwidth-code-point@3.0.0
    │ │ └─┬ strip-ansi@6.0.1
    │ │   └── ansi-regex@5.0.1
    │ ├── type-fest@0.20.2
    │ ├─┬ widest-line@3.1.0
    │ │ └── string-width@4.2.3 deduped
    │ └─┬ wrap-ansi@7.0.0
    │   ├── ansi-styles@4.3.0 deduped
    │   ├── string-width@4.2.3 deduped
    │   └── strip-ansi@6.0.1 deduped
    ├─┬ chalk@4.1.2
    │ ├─┬ ansi-styles@4.3.0
    │ │ └─┬ color-convert@2.0.1
    │ │   └── color-name@1.1.4
    │ └─┬ supports-color@7.2.0
    │   └── has-flag@4.0.0
    ├─┬ configstore@5.0.1
    │ ├─┬ dot-prop@5.3.0
    │ │ └── is-obj@2.0.0
    │ ├── graceful-fs@4.2.8
    │ ├─┬ make-dir@3.1.0
    │ │ └── semver@6.3.0
    │ ├─┬ unique-string@2.0.0
    │ │ └── crypto-random-string@2.0.0
    │ ├─┬ write-file-atomic@3.0.3
    │ │ ├── imurmurhash@0.1.4
    │ │ ├── is-typedarray@1.0.0
    │ │ ├── signal-exit@3.0.6
    │ │ └─┬ typedarray-to-buffer@3.1.5
    │ │   └── is-typedarray@1.0.0 deduped
    │ └── xdg-basedir@4.0.0 deduped
    ├── has-yarn@2.1.0
    ├── import-lazy@2.1.0
    ├─┬ is-ci@2.0.0
    │ └── ci-info@2.0.0
    ├─┬ is-installed-globally@0.4.0
    │ ├─┬ global-dirs@3.0.0
    │ │ └── ini@2.0.0
    │ └── is-path-inside@3.0.3
    ├── is-npm@5.0.0
    ├── is-yarn-global@0.3.0
    ├─┬ latest-version@5.1.0
    │ └─┬ package-json@6.5.0
    │   ├─┬ got@9.6.0
    │   │ ├── @sindresorhus/is@0.14.0
    │   │ ├─┬ @szmarczak/http-timer@1.1.2
    │   │ │ └── defer-to-connect@1.1.3
    │   │ ├─┬ cacheable-request@6.1.0
    │   │ │ ├─┬ clone-response@1.0.2
    │   │ │ │ └── mimic-response@1.0.1 deduped
    │   │ │ ├─┬ get-stream@5.2.0
    │   │ │ │ └── pump@3.0.0 deduped
    │   │ │ ├── http-cache-semantics@4.1.0
    │   │ │ ├─┬ keyv@3.1.0
    │   │ │ │ └── json-buffer@3.0.0
    │   │ │ ├── lowercase-keys@2.0.0
    │   │ │ ├── normalize-url@4.5.1
    │   │ │ └─┬ responselike@1.0.2
    │   │ │   └── lowercase-keys@1.0.1 deduped
    │   │ ├─┬ decompress-response@3.3.0
    │   │ │ └── mimic-response@1.0.1 deduped
    │   │ ├── duplexer3@0.1.4
    │   │ ├─┬ get-stream@4.1.0
    │   │ │ └─┬ pump@3.0.0
    │   │ │   ├─┬ end-of-stream@1.4.4
    │   │ │   │ └── once@1.4.0 deduped
    │   │ │   └─┬ once@1.4.0
    │   │ │     └── wrappy@1.0.2
    │   │ ├── lowercase-keys@1.0.1
    │   │ ├── mimic-response@1.0.1
    │   │ ├── p-cancelable@1.1.0
    │   │ ├── to-readable-stream@1.0.0
    │   │ └─┬ url-parse-lax@3.0.0
    │   │   └── prepend-http@2.0.0
    │   ├─┬ registry-auth-token@4.2.1
    │   │ └─┬ rc@1.2.8
    │   │   ├── deep-extend@0.6.0
    │   │   ├── ini@1.3.8
    │   │   ├── minimist@1.2.5
    │   │   └── strip-json-comments@2.0.1
    │   ├─┬ registry-url@5.1.0
    │   │ └── rc@1.2.8 deduped
    │   └── semver@6.3.0
    ├─┬ pupa@2.1.1
    │ └── escape-goat@2.1.1
    ├─┬ semver-diff@3.1.1
    │ └── semver@6.3.0
    ├─┬ semver@7.3.5
    │ └─┬ lru-cache@6.0.0
    │   └── yallist@4.0.0
    └── xdg-basedir@4.0.0
```
`update-notifier` pulls in more dependencies than the rest of `nodemon`.

I personally think that the drawbacks from keeping outweigh the benefits:
1. The attack surface is increased, and having many dependencies makes it harder to vet the `nodemon` dependency tree.
    - Over the years, several vulnerability reports for `nodemon` dependencies have come from the `update-notifier` tree. Not all of the vulnerabilities affect `nodemon`, but again, it's not easy to immediately tell.
    - Dependency chain attacks are becoming more common.
    - Currently, `npm outdated` reports that about 70 requirements are at least 1 major version behind. I'm sure that some of the updates are not currently necessary, but perhaps they may soon.
2. In the ~8 years since `update-notifier` was added, the ecosystem has improved visibility and handling of dependency updates. We have `npm outdated` and `yarn upgrade-interactive`, and `npx nodemon` without an explicit dependency will try to run the latest version by default.

Thoughts?

### LINKED ISSUE #2028: Update `package-json` to >=8.0.0 for vulnerability in `got` >= 12.0.0, < 12.1.0, < 11.8.5
### Environment
- `nodemon -v`: 2.0.15
- `node -v`: v14.18.1
- Operating system/terminal environment: macOS 12.5

### Issue
```
nodemon@2.0.15 requires got@^9.6.0 via a transitive dependency on package-json@6.5.0
```
- **Vulnerabilty:** The got package before 12.1.0 for Node.js allows a redirect to a UNIX socket.
- **Affected Versions:** >= 12.0.0, < 12.1.0, < 11.8.5
- **Patched versions:** 12.1.0, 11.8.5
- **References**
  - https://github.com/advisories/GHSA-pfrx-2q88-qq97/
  - https://nvd.nist.gov/vuln/detail/CVE-2022-33987
  - https://github.com/sindresorhus/got/pull/2047
  - https://github.com/sindresorhus/got/compare/v12.0.3...v12.1.0
  - https://github.com/sindresorhus/got/commit/861ccd9ac2237df762a9e2beed7edd88c60782dc
  - https://github.com/sindresorhus/got/releases/tag/v11.8.5
  - https://github.com/sindresorhus/got/releases/tag/v12.1.0

### Possible fix
- Upgrade dependency version of `package-json` to >=8.0.0 as it points to fixed dependency for `got` >12.1.0

alexbrazier:
- Closes #1961
- Closes #2028
- Fixes security issue with `got` (CVE-2022-33987)
- Replace `update-notifier` with `simple-update-notifier` which does the same thing but has one dependency (`semver`) rather than several
- Same caching settings as `update-notifier`

## Demo

image

### DIFF
--- a/bin/nodemon.js
+++ b/bin/nodemon.js
@@ -12,5 +12,5 @@ const fs = require('fs');
 const pkg = JSON.parse(fs.readFileSync(__dirname + '/../package.json'));
 
 if (pkg.version.indexOf('0.0.0') !== 0 && options.noUpdateNotifier !== true) {
-  require('update-notifier')({ pkg }).notify();
+  require('simple-update-notifier')({ pkg });
 }
--- a/package-lock.json
+++ b/package-lock.json
@@ -1129,19 +1129,6 @@
         }
       }
     },
-    "@sindresorhus/is": {
-      "version": "0.14.0",
-      "resolved": "https://registry.npmjs.org/@sindresorhus/is/-/is-0.14.0.tgz",
-      "integrity": "sha512-9NET910DNaIPngYnLLPeg+Ogzqsi9uM4mSboU5y6p8S5DzMTVEsJZrawi+BoDNUVBa2DhJqQYUFvMDfgU062LQ=="
-    },
-    "@szmarczak/http-timer": {
-      "version": "1.1.2",
-      "resolved": "https://registry.npmjs.org/@szmarczak/http-timer/-/http-timer-1.1.2.tgz",
-      "integrity": "sha512-XIB2XbzHTN6ieIjfIMV9hlVcfPU26s2vafYWQcZHWXHOxiaRZYEDKEwdl129Zyg50+foYV2jCgtrqSA6qNuNSA==",
-      "requires": {
-        "defer-to-connect": "^1.0.1"
-      }
-    },
     "@tootallnate/once": {
       "version": "2.0.0",
       "resolved": "https://registry.npmjs.org/@tootallnate/once/-/once-2.0.0.tgz",
@@ -1247,14 +1234,6 @@
         "uri-js": "^4.2.2"
       }
     },
-    "ansi-align": {
-      "version": "3.0.1",
-      "resolved": "https://registry.npmjs.org/ansi-align/-/ansi-align-3.0.1.tgz",
-      "integrity": "sha512-IOfwwBF5iczOjp/WeY4YxyjqAFMQoZufdQWDd19SEExbVLNXqvpzSJ/M7Za4/sCPmQ0+GRquoA7bGcINcxew6w==",
-      "requires": {
-        "string-width": "^4.1.0"
-      }
-    },
     "ansi-colors": {
       "version": "4.1.3",
       "resolved": "https://registry.npmjs.org/ansi-colors/-/ansi-colors-4.1.3.tgz",
@@ -1281,15 +1260,8 @@
     "ansi-regex": {
       "version": "5.0.1",
       "resolved": "https://registry.npmjs.org/ansi-regex/-/ansi-regex-5.0.1.tgz",
-      "integrity": "sha512-quJQXlTSUGL2LH9SUXo8VwsY4soanhgo6LNSm84E1LBcE8s3O0wpdiRzyR9z/ZZJMlMWv37qOOb9pdJlMUEKFQ=="
-    },
-    "ansi-styles": {
-      "version": "4.3.0",
-      "resolved": "https://registry.npmjs.org/ansi-styles/-/ansi-styles-4.3.0.tgz",
-      "integrity": "sha512-zbB9rCJAT1rbjiVDb2hqKFHNYLxgtk8NURxZ3IZwD3F6NtxbXZQCnnSi1Lkx+IDohdPlFp222wVALIheZJQSEg==",
-      "requires": {
-        "color-convert": "^2.0.1"
-      }
+      "integrity": "sha512-quJQXlTSUGL2LH9SUXo8VwsY4soanhgo6LNSm84E1LBcE8s3O0wpdiRzyR9z/ZZJMlMWv37qOOb9pdJlMUEKFQ==",
+      "dev": true
     },
     "ansicolors": {
       "version": "0.3.2",
@@ -1394,21 +1366,6 @@
       "integrity": "sha512-VHiNCbI1lKdl44tGrhNfU3lup0Tj/ZBMJB5/2ZbNXRCPuRCO7ed2mgcK4r17y+KB2EfuYuRaVlwNbAeaWGSpbw==",
       "dev": true
     },
-    "boxen": {
-      "version": "5.1.2",
-      "resolved": "https://registry.npmjs.org/boxen/-/boxen-5.1.2.tgz",
-      "integrity": "sha512-9gYgQKXx+1nP8mP7CzFyaUARhg7D3n1dF/FnErWmu9l6JvGpNUN278h0aSb+QjoiKSWG+iZ3uHrcqk0qrY9RQQ==",
-      "requires": {
-        "ansi-align": "^3.0.0",
-        "camelcase": "^6.2.0",
-        "chalk": "^4.1.0",
-        "cli-boxes": "^2.2.1",
-        "string-width": "^4.2.2",
-        "type-fest": "^0.20.2",
-        "widest-line": "^3.1.0",
-        "wrap-ansi": "^7.0.0"
-      }
-    },
     "brace-expansion": {
       "version": "1.1.11",
       "resolved": "https://registry.npmjs.org/brace-expansion/-/brace-expansion-1.1.11.tgz",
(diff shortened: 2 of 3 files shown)


remy:
this is good, but I'm wary your windows test didn't pass: https://github.com/alexbrazier/simple-update-notifier/runs/7033149767?check_suite_focus=true

alexbrazier:
> this is good, but I'm wary your windows test didn't pass: https://github.com/alexbrazier/simple-update-notifier/runs/7033149767?check_suite_focus=true

@remy thanks for looking into it! I actually force pushed that update as the `isTTY` check was causing the update checker not to run on CI so had to update a flag to force it to run during the tests. That is original test run from the old commit and the latest one has all three passing https://github.com/alexbrazier/simple-update-notifier/actions/runs/2554631152

zorn-v:
Seems it is not compatible with node 8.10.0 (minimum required version for nodemon) because `for await`
https://github.com/alexbrazier/simple-update-notifier/blob/91085738fbd27138dc510069d663e013e81af1eb/src/getDistVersion.ts#L10

alexbrazier:
> Seems it is not compatible with node 8.10.0 (minimum required version for nodemon) because `for await`
> https://github.com/alexbrazier/simple-update-notifier/blob/91085738fbd27138dc510069d663e013e81af1eb/src/getDistVersion.ts#L10

Thanks @zorn-v, I've now pushed an update to fix that

remy:
@alexbrazier worth wrapping this in a try/catch - https://github.com/alexbrazier/simple-update-notifier/blob/master/src/getDistVersion.ts#L13= - I can't remember what happens if this fails inside of the deep event handler - might be worth injecting some bad code to see how it behaves.

alexbrazier:
> @alexbrazier worth wrapping this in a try/catch - https://github.com/alexbrazier/simple-update-notifier/blob/master/src/getDistVersion.ts#L13= - I can't remember what happens if this fails inside of the deep event handler - might be worth injecting some bad code to see how it behaves.

Sure, something like this? https://github.com/alexbrazier/simple-update-notifier/pull/2/files

MaximMaximS:
Tbh, you should probably choose this or #2035, because it is pretty bad to have a random vulnerability in the project, which is used by a lot of people. When using npm, the alert is kinda discouraging.

oldium:
@alexbrazier have you thought about using semver module instead of implementing your own version check?

alexbrazier:
> @alexbrazier have you thought about using semver module instead of implementing your own version check?

I was originally trying to keep it simple but now the version check has become more complicated than expected and still not perfect so happy to swap it out for `semver` if that is preferred?

remy:
> Tbh @remy, you should probably choose this or #2035, 

I'm letting @alexbrazier have some time to settle the code base. It's not a straight up simple change and a few extra days to help iron out issues will be worthwhile in the long run. 

Plus, the reality of the vuln in nodemon has near zero impact. Of course it's not zero, but there's no path to exploit the actual vuln through nodemon (partly because it's fired outside of calling your code - which would take advantage of the vuln, and partly because nodemon in a lot of cases forks the sub process, so it never has access to any of the nodemon code - and thusly the vuln on `got`).

The issue isn't being ignored, @alexbrazier has kindly and valiantly taken up the mantle to solve this dependency once and for all.

alexbrazier:
Thanks @remy!

As the last couple of changes have been around some issues with the semver comparison and there were still some edge cases I've decided to just switch to using [semver](https://github.com/npm/node-semver) to make it more reliable. Should hopefully be good to go now unless anyone has any other comments.

zorn-v:
semver 7.3.7 has
```json
  "engines": {
    "node": ">=10"
  }
```

oldium:
nodemon is using this version: "semver": "^5.7.1"

alexbrazier:
> semver 7.3.7 has
> 
> ```json
>   "engines": {
>     "node": ">=10"
>   }
> ```
> 
> There is `"semver": "^5.7.1"` in nodemon deps

Hmm, I've tested with node `8.10.0` and it seems to work as expected:

image

remy:
> > semver 7.3.7 has
> > ```json
> >   "engines": {
> >     "node": ">=10"
> >   }
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > There is `"semver": "^5.7.1"` in nodemon deps
> 
> Hmm, I've tested with node `8.10.0` and it seems to work as expected:
> 
> image

Are there any warnings during install if you're using node 8? It could be just that.

alexbrazier:
> Are there any warnings during install if you're using node 8? It could be just that.

Nothing when installing with npm, but seems there is an [issue with yarn](https://github.com/npm/node-semver/commit/d61f828e64260a0a097f26210f5500e91a621828#r36473304)
image

Axent96:
related to https://github.com/remy/nodemon/issues/2040

zorn-v:
> > Are there any warnings during install if you're using node 8? It could be just that.
> 
> Nothing when installing with npm, but seems there is an [issue with yarn](https://github.com/npm/node-semver/commit/d61f828e64260a0a097f26210f5500e91a621828#r36473304) image

You can use `~7.0.0` to avoid then

slaytr:
hi @remy is there a projected release date for this change?

remy:
@alexbrazier I want to merge this up, I feel like it's had it's time to percolate. But when I merge there's suddenly going to be, quite literally, millions of users in an absolute boat load of different scenarios using this library through nodemon.

So what I wanted to ask is: is there *anything* you're not 100% sure about? (Not meant as *this is all on you*!). I've been reading and re-reading your code to see if I can see anything. 

Couple of things I wondered were:

- what happens with a network timeout (though IIRC these are really hard to catch as they seem to fire outside of the promises...)
- is it possible that the config directory or files can't be written, and what exception is thrown?
- are there any environments that original update-notifier disabled itself (like CI or TEST) that yours should take inspiration from (I currently don't think there are, but just putting the question out).

alexbrazier:
@remy I would say I'm fairly confident with it as it's been tested down to node `8.10.0` and it currently handles errors in the background and chooses not to output anything if this happens to be on the safe side. I can't be 100% sure the cache/config dir is going to work on every OS particularly if there are some strict file permissions, however it's set to just catch these errors and just ignore the update check for now so should be handled.

I think it's the same thing for network timeouts or errors (I've tested without an internet connection) where the promise will reject it will just catch the error.

For the testing scenario it uses `process.stdout.isTTY` so will be ignored on CI/scripts which I think should cover most cases and I think is the same as what the original `update-notifier` used.

peter-sattler:
Thanks for fixing this @alexbrazier.  Looking forward to its deployment. :)

jimmywarting:
Don't really like that modules creates files outside of it's own directory and caches things in config store files...
when i uninstall something then i want this files to be removed as well...

remy:
@jimmywarting update-notifier was always doing this, and if you look in your `~/.config` directory (assuming mac or linux - not sure where it is on windows), you'll find the other systems that store files outside of the directory.

oldium:
What about this project focused to standardize the way of using the .cache folder? https://github.com/avajs/find-cache-dir

remy:
Firstly, it isn't standardised. Secondly, again, the aim was to keep things the same as possible, and update-notifier was using an existing .config directory for many years prior.

peter-sattler:
@jimmywarting and @remy - My suggestion is to create a different story if this makes sense to do, then the work can be prioritized accordingly. It’s not really a part of this PR.

PR #2052, opened 2022-07-14
chore: upgrade minimatch dependency to 3.1.2

goliney:
I know it's just a formality, but our customers are obsessed with audit logs and they are going crazy about vulnerability in `minimatch@3.0.4`.


goliney:
@remy I'm sorry to bug you with this, I really do, but do you happen to know when we can expect a release of a new `nodemon` version with the upgraded `minimatch` dependency?

PR #2059, opened 2022-08-26
Fix for : Giving preference to "main" of package.json for script in nodemon command

### LINKED ISSUE #2051: Argumentless nodemon command in npm project does not run value of "main" from package.json
### Expected behaviour
See stackoverflow question: https://stackoverflow.com/questions/72946313/

npm project js file can be run by the nodemon command without arguments as long as "main" set in package.json file. For example, if main was set to "server.js", when I run "nodemon", then what happens should be equivalent to "nodemon server.js". It shouldn't matter if there is an "index.js" file at the root of my project, the "nodemon" command should still run "nodemon server.js" 

### Actual behaviour
When there is an "index.js" file at the root of my project, it doesn't matter what the "main" attribute is in "package.json": it will always try to run "index.js" at root when I run "nodemon".

### Steps to reproduce
Make a directory
cd to that directory and make a new npm project with 'npm init -y'
in the package.json file, set main to "server.js"
make a server.js file at the root of the project, have it filled with some simple code, perhaps a simple Express app
make a index.js file at root, have it filled with some simple code
run `nodemon`

### Environment
nodemon 2.0.16
macOS Monterey 12.4
zsh 5.8.1 (x86_64-apple-darwin21.0)
npm 1.5.5
node 16.15.0

Vindeep07:
Fixes #2051 
Added fix for giving preference `main` of package.json for setting script in execOptions of nodemon command.

- Instead of using `index.js` directly for `script` in `configOptions`, check for `main` in package.json and if `main` does not exists then use `index.js`.


remy:
Not intentionally sitting on this - just been busy with IRL work! :) I'll try my best to review this fully next week and get it live.

Thank you 🙏

remy:
lol, "next week" - suuuuure Rem…

Looking again now.

remy:
So, wondering if this is a bug fix or if it's a breaking change of current behaviour.

Certainly it fixes how I intended it to work… 🤔 

I'm just not keen on falling into breaking change (which is a major version bump) if possible (i.e. it's easier for people to adopt a bug fix change in semver).

voxpelli:
> So, wondering if this is a bug fix or if it's a breaking change of current behaviour.
> 
> Certainly it fixes how I intended it to work… 🤔
> 
> I'm just not keen on falling into breaking change (which is a major version bump) if possible (i.e. it's easier for people to adopt a bug fix change in semver).

Could you publish a version where it prints a warning when `"main"` will not result in a found `index.js`? And then a month later or so ship the change?

remy:
Since I needed to drop node@8 support due to semver vuln, I've included this in the major bump

PR #2073, opened 2022-10-16
Update style.css

parassingh016:
change in padding


remy:
I guess, why? 

I also intentionally chose steps of 4px...

PR #2098, opened 2023-02-01
Get rid of spawning shell windows if nodemon is started without console.

lerarosalene:
If nodemon is started without it's own console (example: via pm2) on Windows, it spawns console windows for it's monitored processes. This should get rid of this behaviour.


lerarosalene:
Related issue: https://github.com/remy/nodemon/issues/2099

PR #2102, opened 2023-03-06
newly updated nodemon website with react

ayushkavad:
newly updated nodemon website


ayushkavad:
this is newly updated nodemon website with documentation build with React.
nodemon

ayushkavad:
nodemon2

ayushkavad:
nodemon3

remy:
Sorry chap, but the homepage is a single static html file which is a million times better than these new 37 files. Closing for that reason.

client/src/index.js
0123sandeepkumar, line 1: what is component doing

PR #2115, opened 2023-05-25
Update README.md: Added explanantion about locally and globally insta…

alirazaNaqvi5:
…lled problem.

Most of the new developers face this problem, and they don't know how to tackle it.

so I have added an explanation, in usage section, that if they used npm command without "-g" flag then it is installed locally in that proiect, and they won't be able to use "nodemon" directly in terminal.

### DIFF
--- a/README.md
+++ b/README.md
@@ -31,6 +31,20 @@ With a local installation, nodemon will not be available in your system path or 
 
 # Usage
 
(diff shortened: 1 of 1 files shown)


remy:
Closing this as I'm fairly confident that it's covered as well as it can be right at the start of the existing readme.

PR #2116, opened 2023-06-13
Replace Travis badge with GitHub Actions badge

kyletaylored:
It doesn't look like Travis CI is being used anymore and that GitHub Actions is being used for various tasks, so this just replaces the tests badge on the README.

Unfortunately the current test workflow (`node.js.yml`) runs on both PRs and merges to `main`, so if a PR someone opens up fails, the badge also reflects that, so ideally there would be two separate workflows to break out those responsibilities.


bitcoinfinancier (approved):

remy:
I've removed this entirely for the time being - thanks for the heads up about how github works 👍

Issue #19, opened 2011-04-01
nodemon-ignore vs .nodemonignore


timoxley:
Many other 'meta' files that are generated by IDEs and tools like git/mercurial, are .hidden by default, as they aren't actually part of the project, they're part of the files supporting the project.

To keep a degree of consistency with these other systems, and so I can easily cluster all of my meta files together/hide them, it'd be good if nodemon's ignore file was .nodemonignore by default, dropping the hypen also keeps it consistent with the naming format of .gitignore/.hgignore.

to ensure compatibility with existing projects nodemon could auto-rename 'legacy' nodemon-ignore files to .nodemonignore

PR #2118, opened 2023-06-23
Address vuln in semver (bump simple-update-notifier & semver)

Primexz:
fix some vulns

### DIFF
--- a/package.json
+++ b/package.json
@@ -9,7 +9,7 @@
     "nodemon": "./bin/nodemon.js"
   },
   "engines": {
-    "node": ">=8.10.0"
+    "node": ">=10"
   },
   "repository": {
     "type": "git",
@@ -61,7 +61,7 @@
     "minimatch": "^3.1.2",
     "pstree.remy": "^1.1.8",
     "semver": "^7.5.3",
-    "simple-update-notifier": "^1.1.0",
+    "simple-update-notifier": "^2.0.0",
     "supports-color": "^5.5.0",
     "touch": "^3.1.0",
     "undefsafe": "^2.0.5"
--- a/package.json
+++ b/package.json
@@ -60,8 +60,8 @@
     "ignore-by-default": "^1.0.1",
     "minimatch": "^3.1.2",
     "pstree.remy": "^1.1.8",
-    "semver": "^5.7.1",
-    "simple-update-notifier": "^1.0.7",
+    "semver": "^7.5.3",
+    "simple-update-notifier": "^1.1.0",
     "supports-color": "^5.5.0",
     "touch": "^3.1.0",
     "undefsafe": "^2.0.5"
--- a/package-lock.json
+++ b/package-lock.json
@@ -15,7 +15,7 @@
         "minimatch": "^3.1.2",
         "pstree.remy": "^1.1.8",
         "semver": "^7.5.3",
-        "simple-update-notifier": "^1.1.0",
+        "simple-update-notifier": "^2.0.0",
         "supports-color": "^5.5.0",
         "touch": "^3.1.0",
         "undefsafe": "^2.0.5"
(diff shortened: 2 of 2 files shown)


FBNitro:
Not an issue for me, but... semver 7.x needs node 10+.  The nodemon package.json says:
  "engines": {
    "node": ">=8.10.0"
  },
  
Which means this could be a breaking change for 'someone' out there.

aren55555:
Support for Node 8 ended 3 years and 5 months ago (31 Dec 2019). Even support for Node 10 ended 2 years ago (30 Apr 2021). See https://endoflife.date/nodejs

The `package.json` should be updated to something a lot more recent, and hopefully at least in the "Security Support" phase. It doesn't do anyone any good to support versions of Node that are no longer maintained and potentially insecure.

wellwelwel:
@Primexz, these change will not fix the problem since the `simple-update-notifier` dependency has the same problem.
> See the [simple-update-notifier PR #19](https://github.com/alexbrazier/simple-update-notifier/pull/19)

The `nodemon` depends on this being fixed in the `simple-update-notifier` first, probably in a version `^2.0.0`.

@aren55555, regardless of whether it's an unsupported version of NodeJS, following the release semantics,  it would probably be the ideal way to release a major version of `nodemon` like `^3.0.0`.

Then, indicate the support starting at Node 10 in:
https://github.com/remy/nodemon/blob/6787871d521eef65c2bc7a62234e3736bf6fcc35/package.json#L11-L13

Primexz:
Then we wait until a fix is merged into the master of simple-update-notifier.
If a fix is available there we can continue with this PR.

remy:
Firstly, sorry for late replies. Gmail has been slurping a lot of legit emails into spam, so I've missed a lot (including replies to emails I've sent my own family).

Regarding support for older node versions that @aren55555 mentions, I'll explain: node 8 was easy to support for a long time with little to no changes. Since I come from the web and backward compatibility is the gold standard, *and* when it was easy to do, I did. 

I'd only drop an older version of node of there was a really good reason (like the version of node flat out couldn't support some feature).

I'm happy to drop node 8 if semver simply can't support it, but I'll only go to the lowest common version (which I'd expect is 10, but again, happy to push higher if it's required).

Some people are in companies that can't easily upgrade, and it's important to me to try to help them as long as I can. 

Aside, I'm sure the author of the updater script will be happy to merge a pr - the module development was driven by stale support in the original updater and nodemon relying on it.

(Completely aside, I'd love to know what changed in semver that meant they had to drop support for node 8 - I'm guessing native await support, but I'll have to take a gander).

Thanks to everyone bringing this up.

wellwelwel:
I think it's can be good:

I tested the most common methods from `semver`, then I got "breaks" on Node 8 using:
- semver.simplifyRange;
- semver.premajor;
- semver.preminor;
- semver.prepatch;
- semver.literals;

But when I use any other method, every line works fine (including the `satisfies`, the only method used by `nodemon` and `gt`, the only method used in `simple-update-notifier`).

Since the `nodemon` don't uses any method that can breaks on Node 8, I think it doesn't need a break change.

@remy, I would like your opinion about that.

alexbrazier:
Version [2.0.0](https://github.com/alexbrazier/simple-update-notifier/releases/tag/v2.0.0) of `simple-update-notifier` has just been release which updates to the latest `semver` versions.

It is a major version update because it officially drops support for node 8, however it still appears to work with node 8, but you'll get an error when running `yarn install` on node 8.

For this PR you might also want to update the package.json to `"node": ">=10"` to match `simple-update-notifier` and `semver`.

## Node 8 workaround

Run yarn with `yarn install --ignore-engines` to ignore node 8 error and it should still work. I don't think you get an error when using `npm install`.

package.json
```diff
@@ -60,8 +60,8 @@
     "ignore-by-default": "^1.0.1",
     "minimatch": "^3.1.2",
     "pstree.remy": "^1.1.8",
-    "semver": "^5.7.1",
-    "simple-update-notifier": "^1.0.7",
+    "semver": "^7.5.3",
+    "simple-update-notifier": "^1.1.0",
```
JimmyBjorklund, line 64: Bump this to 2.0.0 to get rid of semver problem.

Primexz: Will do that soon

wellwelwel: @JimmyBjorklund, shouldn't the `simple-update-notifier` version be `^2.0.0`?

wellwelwel (approved):

wellwelwel:
Thanks @Primexz 🚀

SP-JohannT:
> Thanks @Primexz 🚀

np :)

package.json
JimmyBjorklund: You are right it should

JimmyBjorklund (approved):

chinwahc (approved):

Cellule:
Can we get this merged and published ? 
Would love to get rid of semver's vulnerability warnings in my project

remy:
I've merged this - along with some other changes at the same time. Tests are doing something weird, but I've validated them manually and I'm happy.

The whole CI system is a little brittle (due to the way the tests need to constantly spawn processes), but if it [releases](https://github.com/remy/nodemon/actions/runs/5484732578/jobs/9992666966) you should see this in nodemon@3. The first major release in a long time (and hopefully the last, I really don't like breaking changes, but it absolutely makes sense to finally drop official support for node@8).

PR #2133, opened 2023-09-20
Add node 20 to test matrix

BackEndTea:


didaquis (approved):

Abuchtela (commented):
commit 32807b56eae8392111c54102220afccf59a0e1c5

Abuchtela (changes requested):
gh pr checkout 2133

Abuchtela (approved):
- [ ] .github/workflows/node.js.yml

remy:
I don't think the tests are failing because of node 20 - but because something's either wrong in the test or there's a regression somewhere…

I'm trying to have a look but don't have a lot of time (on the computer) atm - but if no one else does, I will do and I will get node 20 in the matrix.

jimmywarting:
How would you feel about just testing only the latest 3 versions? 16, 18, 20? and setting minium node version to it as well in package.json
Could start using newer NodeJS features by doing so.

ravin00 (changes requested):
Remove the node 10.x and add the 12.x support and upgrade the suppoted version to 18.x and 20.x. first of all do the checking first before the changes.

remy:
> Remove the node 10.x and add the 12.x support and upgrade the suppoted version to 18.x and 20.x. first of all do the checking first before the changes.

Why are you arbitrarily proposing to drop support for 10 and 12? (This is rhetorical actually and rather it's naive to assume that I'd want that)

ravin00:
> > Remove the node 10.x and add the 12.x support and upgrade the suppoted version to 18.x and 20.x. first of all do the checking first before the changes.
> 
> Why are you arbitrarily proposing to drop support for 10 and 12? (This is rhetorical actually and rather it's naive to assume that I'd want that)

proposed that because of the security purposes i told that to change the version to 18.x and 20.x be the ones there do not receive any security updates

remy:
On reflection, it would be ideal to _only_ run integration tests on the lowest common node version, rather than the latest with all the bells.

ravin00:
> On reflection, it would be ideal to _only_ run integration tests on the lowest common node version, rather than the latest with all the bells.

If that's the case your correct

PR #2204, opened 2024-05-19
fix: add types to help with required nodemon usage

remy:
Hopefully helps devs use nodemon as required and adds some advice in tools like VS Code.


chuyjr76033 (commented):
🤑

Issue #2206, opened 2024-05-28
Problem with Newley released types


KevinGruber:
- Versions: node@v22.1.0, darwin@23.2.0
- `nodemon -v`: 3.1.1
- Operating system/terminal environment (powershell, gitshell, etc):  MacOs, terminal
- Command you ran: Via api in typescript

### Expected behaviour
When importing nodemon there is a default export for functionality defined in the types.

### Actual behaviour
With the addition of https://github.com/remy/nodemon/pull/2204 there where types added to the project, but no default export types for functionality -> that leads to a not usable nodemon in typescript

### Steps to reproduce
import nodemon from 'nodemon';

->  nodemon is not a module

---

remy:
Just checking, but does require work? And can you import using `import * as nodemon from 'nodemon'?

KevinGruber:
const nodemon = require('nodemon'); - WORKS doesn't load types

import * as nodemon from 'nodemon'; - DOESN't work

I am pretty sure it is because there is no export of the functionality of nodemon.

Based on the types nothing is exported

remy:
I need to have a play. I only added types as an attempt to help people who were requiring the module, but I don't personally use TypeScript ([some context](https://remysharp.com/2024/02/23/why-my-code-isnt-in-typescript)). If you have any advice, please do share, otherwise I'll tinker this week 👍

remy:
Okay, to clarify, `import nodemon from 'nodemon'` does work, i.e. this code works:

```
nodemon({}).on('start', () => {
  console.log('started');
});
```

But the types aren't being found by tools like VS Code.

This is my code:

package.json

```json
{
  "name": "2206",
  "version": "1.0.0",
  "main": "index.js",
  "dependencies": {
    "nodemon": "^3.1.1"
  }
}
```

index.js (the sub process):

```js
console.log('Hello, World!');
```

And the run script: test.mjs

```js
import nodemon from 'nodemon';

nodemon({}).on('start', () => {
  console.log('started');
});
```

From the command line:

<img width="687" alt="SCR-20240528-jwdv" src="https://github.com/remy/nodemon/assets/13700/6fe4caf9-531a-4581-8c96-8073d5991a65">

@KevinGruber you're saying it's undefined, can you share your code?

KevinGruber:
Ok sry for the misunderstanding.

The runtime code works yes, but the typescript doesn't build because of "wrong" types.

undefined in the sense of types in TS

[...]

PR #2207, opened 2024-05-28
fix: Type exports correctly

KevinGruber:
As per discussion in: https://github.com/remy/nodemon/issues/2206

feat(types): export types
feat(types): add default export

@remy I added two TODO's with questions. Also the TEMP_TEST_TS file is just there for you to easier test it locally, I will remove it before merging.

BR,
Kev

### DIFF
--- a/TEMP_TEST_TS.ts
+++ b/TEMP_TEST_TS.ts
@@ -1,0 +1,5 @@
+import nodemon from './index';
+
+const test = nodemon({
+    script: ''
+});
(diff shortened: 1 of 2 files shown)


remy:
Looks good to me. If you can remove the tmp file, I'll get it merged it.

remy:
I'll sort out the conventional commitments failure, it's not exactly useful in its error...

remy:
@KevinGruber thanks again for this - I've merged it as a bug fix rather than feature (which the last introduction of types did). Cheers for flagging.

PR #2211, opened 2024-06-21
Fix types for export on ESModule

ani3llyon:
Fix types for export in file index.d.ts

- use typescript
- set `"type": "module"` in package.json or set extention file to .mts
```
import nodemon from 'nodemon';

// no show error on ide and also when compile the code
// but error when running
nodemon.default();

// show error on ide and also when compile the code
// but no error when running
nodemon();
```


akantic:
This error is present again in 3.1.9.

PR #2235, opened 2024-11-07
chore: bump chokidar to v4

zjbob:
Chokidar v4 removed a bunch of its dependencies.
Implicit dependecies, such as braces, have critical issues.

   - braces's issue is particular is: Uncontrolled resource consumption

This issue was brought to my attention from Dependabot in my project and is the reason I am proposing this change.

Most notable is that [Chokidar v4](https://github.com/paulmillr/chokidar/releases/tag/4.0.0) reduced its dependencies from 13 to 1.

Updating Chokidar implicitly changes the requirement of supported node to >=14.16.0 since it is a requirement of Chokidar's now only dependency:

   - readdirp
  
Due to the minimum requirement of node being changed to 14.6, I believe that this change requires a nodemon major version bump. I will defer this decision and action to the repo maintainer.

_Following are edits by @remy_

- [ ] Tested on MacOS
- [ ] Tested on Windows cmd
- [ ] Tested on Windows powershell
- [ ] Tested on Windows WSL (jeez, this list goes on)
- [ ] Tested on linux

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -9,7 +9,7 @@
       "version": "0.0.0-development",
       "license": "MIT",
       "dependencies": {
-        "chokidar": "^3.5.2",
+        "chokidar": "^4.0.1",
         "debug": "^4",
         "ignore-by-default": "^1.0.1",
         "minimatch": "^3.1.2",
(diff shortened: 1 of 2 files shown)


zjbob:
@remy When you have a spare moment, curious if you have any thoughts

remy:
I've not run tests on this yet, but I'm wary of creating a major version bump (because support is reduced).

Nodemon doesn't have any telemetry (though there is an updater) so I can't tell with any degree at all how many devs are still on node@12. 

Just adding this comment so people understand where I'm coming from. I'll do some digging into the pr to understand the change that Chokidar brings and where the lost support is.

remy:
Related issue on Chokiar to see if node@12 can be supported through polyfill: https://github.com/paulmillr/chokidar/issues/1390

benmccann:
> Note for testing - CI fails on this test file: test/monitor/count.test.js

I think there might be a couple of other test failures as well. I looked at them briefly. I think that the issue is that the nodemon watcher ignore uses glob patterns and glob support was dropped in chokidar 4 in favor of a filter function. If those glob patterns can be replaced that would be the best solution.

Axorax (approved):

der-eismann:
@remy any news on that? We're getting hit with security alerts as well because of transient dependencies like picomatch. Can you please update chokidar to v4 or even v5?

remy:
> @remy any news on that? We're getting hit with security alerts as well because of transient dependencies like picomatch. Can you please update chokidar to v4 or even v5?

It's a fully breaking change, requiring me to jump to `3.x` - which I'm wary of doing. I've not had the time (at all) to review the code in Chokidar to find whether it would force me to also drop much older node support.

It's still on my radar, but harder to justify.

Issue #2222, opened 2024-09-15
Upgrade to chokidar 4


benmccann:
Chokidar v4 is out now. It drops a dozen dependencies, so is much smaller and lighter

remy:
I need to check this fairly carefully for any breaking changes. If it's all good, then I will do, but there's no high priority to bump the dep just yet (especially as it's _just_ come out).

benmccann:
The big breaking change is that it dropped support for glob syntax, so any glob usage needs to be handled on the user's side: https://github.com/paulmillr/chokidar?tab=readme-ov-file#upgrading

benmccann:
Still exists as a to-do

benmccann:
Still exists as a to-do

benmccann:
Still exists as a to-do

lgugla:
Any update here as in chokidar 3.6 there is a dependency of braces package which have a high vulnerability reported in our scans. As in chokidar 4 this dependency is removed and the vulnerability will get auto fixed.

Please update to chokidar 4.x so that vulnerability get fixed in this package.
![image](https://github.com/user-attachments/assets/0b5980b6-a896-4a97-910b-3b8b0d123374)

remy:
@lgugla it's worth doing your reading rather than blindly reporting what vuln auto tools report: https://github.com/remy/nodemon/issues/2203

You're not alone though, often vulns are reported when they're nonsense.

remy:
I'm guessing no one has wanted to contribute to this issue to see if it breaks backward support, so I'll do it when I've got time. 

I'm guessing the only tangible benefit is ticking that security box around the false positive flag on braces, or is there something else that's significantly better? (this will help inform the priority)

benmccann:
As mentioned in the issue description, it also drops a dozen dependencies so results in smaller installs and less supply chain risk

benmccann:
Still exists

benmccann:
I'm going to close this issue as I'm tired of dealing with the stale bot. I've removed `nodemon` from my project instead

remy:
There's a PR that I think came in before yours which I commented on earlier this morning and will be testing (just writing for anyone landing on this issue, appreciate doesn't affect @benmccann anymore, but might be useful for others).

benmccann:
Ah, I hadn't seen that. Here's the PR: https://github.com/remy/nodemon/pull/2235

I left a note on it about the test failures

PR #2239, opened 2024-11-26
chore: chokidar 4

benmccann:
closes https://github.com/remy/nodemon/issues/2222


benmccann:
might need to use `tinyglobby` to match `ignored` patterns

PR #2260, opened 2025-11-09
Update npm link in installation instructions

skyeyeye:


skyeyeye:
@aaronjensen plss check

PR #2262, opened 2025-11-30
fix: replace deprecated substr() with slice() in file matching logic

AkaHarshit:
## Summary

This PR replaces the deprecated `.substr()` method with `.slice()` in the file matching logic of `lib/monitor/watch.js`. The change maintains identical functionality while using the modern, non-deprecated string method that will remain supported in future JavaScript versions.

### Problem
The codebase uses the deprecated `.substr()` method in `lib/monitor/watch.js:190` for file path matching. The `.substr()` method is deprecated and may be removed in future JavaScript versions, which could cause compatibility issues.

### Root Cause
The code was written using `.substr(-length, length)` to extract the last `length` characters from a file path string. While this works correctly, `.substr()` is deprecated in favor of `.slice()`, which provides the same functionality with a cleaner API.

### Solution
Replace `file.substr(-length, length)` with `file.slice(-length)`. Both methods produce identical results:
- `.substr(-length, length)` takes `length` characters starting from `length` characters from the end
- `.slice(-length)` takes from `length` characters from the end to the end of the string (equivalent behavior)

## Before vs After

**Before:**
```javascript
if (file.substr(-length, length) === script) {
  // match found
}
```

**After:**
```javascript
if (file.slice(-length) === script) {
  // match found
}
```

**Behavior**: Identical - both extract the last `length` characters from the string and compare with the script path.

## Testing

- ✅ Linter passes (`npm run lint`)
- ✅ Existing tests pass (33 passing in watch tests)
- ✅ No functional changes - the logic produces identical results

## Why This PR Should Be Merged

1. **Fixes a Real Issue**: Addresses the use of a deprecated API that may cause future compatibility issues
2. **No Breaking Changes**: The change is functionally identical - `.slice(-length)` produces the same result as `.substr(-length, length)`
3. **Small & Scoped**: Single line change in one file, minimal risk
4. **Follows Best Practices**: Uses modern JavaScript standards
5. **Future-Proofs Codebase**: Ensures compatibility with future JavaScript versions
6. **Code Quality**: Improves maintainability by removing deprecated API usage

## Files Changed
- `lib/monitor/watch.js` (1 line changed: line 190)

### DIFF
--- a/lib/monitor/watch.js
+++ b/lib/monitor/watch.js
@@ -187,7 +187,7 @@ function filterAndRestart(files) {
       if (matched.result.length === 0 && script) {
         const length = script.length;
         files.find(file => {
-          if (file.substr(-length, length) === script) {
+          if (file.slice(-length) === script) {
             matched = {
               result: [file],
               total: 1,


remy:
Sorry, appreciate the effort in the PR text, but this a noop change that isn't required.

PR #2283, opened 2026-08-22
fix: don't exit on code 2 from a long-running process

ayushkavad:
Ran into this while working on a project: nodemon would sometimes just quit with the "unhandled exit code (2)" message even though my app had been running fine for a long time.

Took a while to track down. The exit handler uses `Date.now() < config.lastStarted + 500` to decide whether the process died right after startup (the "syntax error" heuristic). But `watch.js` resets `config.lastStarted` on *every* change event, including files that don't match any watch rule and don't trigger a restart. So if the app writes a log file and then happens to exit with code 2 within half a second, nodemon thinks the command itself is broken and stops completely instead of waiting for file changes.

The fix just keeps the real start time in a local variable inside `run()` so the watcher can't clobber it. I left the assignment in `watch.js` alone because `runOnChangeOnly` relies on it.

Also added a test that reproduces the scenario (app runs 1.5s, writes a .txt, exits with 2) - it fails on main and passes with this change. Verified that a script exiting with code 2 immediately on startup still stops nodemon like before.

### DIFF
--- a/lib/monitor/run.js
+++ b/lib/monitor/run.js
@@ -48,6 +48,12 @@ function run(options) {
 
   config.lastStarted = Date.now();
 
+  // capture the start time of *this* child locally: config.lastStarted is
+  // also overwritten by the watcher on every file-change event, so it can't
+  // be trusted to measure how long the child has been running (the exit
+  // handler below uses it to detect an instant startup failure)
+  const childStarted = config.lastStarted;
+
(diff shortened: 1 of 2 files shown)


ayushkavad:
@remy when you get a chance, would appreciate a look at this one. The failure mode is nodemon exiting completely mid-session, which had me confused for a good hour before I found the cause. Small diff, test included.

